PR #4, opened 2017-05-25
adding icon to display with site title

hozefaj:


pages/_document.js
```diff
@@ -23,6 +23,8 @@ export default class MyDocument extends Document {
         <Head>
           <title>glamorous - React component styling solved 💄</title>
           <style dangerouslySetInnerHTML={{__html: this.props.css}}/>
+          <meta name="viewport" content="width=device-width, initial-scale=1" />
+          <link rel="shortcut icon" type="image/x-icon" href="https://raw.githubusercontent.com/paypal/glamorous/master/other/logo/full.png" />
```
kentcdodds: Would probably be better to use an asset that's deployed with the rest of the site. I'm not sure how that works though

hozefaj: Actually to think of it makes sense, to have a `asset` folder as a part of the repo and reference it from there. Does that make sense?

If yes, will update the PR accordingly.

paulmolluzzo: Yes, put it in a directory called `static`. https://github.com/zeit/next.js#static-file-serving-eg-images

hozefaj: Cool...Will make the changes...

Issue #5, opened 2017-05-26
Add Documentation


paulmolluzzo:
There is a URL `/docs` that we'd like to populate with documentation. The documentation that we have now is all piled into the [README of `glamorous`](https://github.com/paypal/glamorous/blob/master/README.md).

Ideally the documentation we store will be in markdown to make it easy or other people to contribute.

Since there is no content just yet, there's a lot of flexibility on how to structure this. It might be worth creating a directory like `docs` and then putting the markdown files in there. Then we can load the markdown files into the `/docs` page.

An idea I have is to split the documentation up into separate files like according to the headers. From there we might have an easier time supporting searching and direct linking to topics through query params.

liadbiz:
@paulmolluzzo yeah, this is what I am planing to do, actually some works have been done.
my plan about /docs stucture: 
+ basics
  + install.md
  + get_started.md
  + core_concepts.md
  + props.md
  + theme.md
  + context.md
 ...
+ guide
  + size.md
  + animation.md
  + media_query.md
  + server_rendering.md
  + helper_function.md
...
+ api
  + api.md

I have made a experimental [repo](https://github.com/liadbiz/glamorous-notes)

@kentcdodds  need you thoughts here! :100:

paulmolluzzo:
@liadbiz please take a look at #9 and let me know your thoughts. Maybe we can get it sorta working with one file, merge it, then you can split it up and get the other files with individual routes and a subnav working?

kentcdodds:
@liadbiz, I like that structure. Though for the `api` it may be useful to split it out into multiple pages because the API is kinda big. But maybe not 🤔

kentcdodds:
Alrighty, so after a lot of trial and error, I got Crowdin set up on the repo. You can see the project [here](https://crowdin.com/project/glamorous-website). [Here](https://www.youtube.com/watch?v=F69lAvAKxE4) is a promo video that explains what crowdin is.

So now we need to break up the `main.md` file into smaller bits for all of the pages that we want and start working on the basics/guide/integrations/etc. as well.

We need all the help we can get! I'll create the files that I think we'll need, then I'll ask people to help fill in the files with content.

kentcdodds:
Ok, I think that we're ready to start getting help from folks. A lot of stuff can be pulled out of [the glamorous README.md](https://github.com/paypal/glamorous/blob/master/README.md) and [the examples](https://github.com/paypal/glamorous/tree/master/examples).

[...]

PR #9, opened 2017-05-26
[WIP] - Add basic docs

paulmolluzzo:
This is a super early WIP for getting documentation into the site via .md files. Right now I'm just shoving the README from `Installation` to the end (basically) into the docs page. This addresses #5.

I'd really like some feedback and thoughts here. Ideally I can get this into a clean enough place for us to merge and then let other people flesh more of it out.

In particular, I'm not super sure that the method I have for including the .md file is correct. I get a flash where the content is missing, but every other method I've tried has failed. I'm not an expert on webpack (or anything for that matter 😭 ).

Looking at this and the documentation in general I feel that we'll wind up wanting to "hardcode" the documentation on the homepage and not use a md file. I think we'll have way more control over the look that way, and since the homepage is probably going to be a brief intro and installation instructions I don't see why we'll need to worry about lots of updates.

Deployed this here: https://glamorous-website-poiyvykcjm.now.sh

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -1,16 +1,31 @@
-// This file will let us add in prefetch conditionally so we don't break jest snapshot testing
-
 const webpack = require('webpack')
+const marked = require('marked')
 
+const renderer = new marked.Renderer()
 const USE_PREFETCH = process.env.NODE_ENV !== 'test'
 
 module.exports = {
   webpack: config => {
+    // Add in prefetch conditionally so we don't break jest snapshot testing
     config.plugins.push(
       new webpack.DefinePlugin({
         'process.env.USE_PREFETCH': JSON.stringify(USE_PREFETCH)
       })
     )
+
+    // Markdown loader so we can use docs as .md files
+    config.module.rules.push({
+      test: /\.md$/,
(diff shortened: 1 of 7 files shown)


next.config.js
```diff
@@ -1,17 +1,32 @@
...
 module.exports = {
   webpack: config => {
+    // Add in prefetch conditionally so we don't break jest snapshot testing
     config.plugins.push(
       new webpack.DefinePlugin({
         'process.env.USE_PREFETCH': JSON.stringify(USE_PREFETCH)
       })
     )
 
+    // Markdown loader so we can use docs as .md files
```
kentcdodds: This could probably be:

```javascript
config.module.rules.push({
  test: /\.md$/,
  use: [
    {loader: 'html-loader'},
    {
      loader: 'markdown-loader',
      options: {pedantic: true, renderer},
    },
  ]
})
```

paulmolluzzo: Done.

paulmolluzzo:
Are we good to merge?

liadbiz:
@paulmolluzzo  made a PR to tweak the style of docs.md. #13

PR #16, opened 2017-05-29
[WIP] Use theme provider

paulmolluzzo:
**What**: This adds in support for the `ThemeProvider` and adds some other base styles.

**Why**: This should make it much easier to manage global/shared styles across components

**How**: Added the `ThemeProvider` as a wrapping component in `layout.js`

This is WIP really, and all the components that are using values from  `colors.js` can and should be updated to use the `theme` and pull values from there. If we get impatient and this is merged before I get back to this then that's refactoring that will be out there for someone to tackle.


paulmolluzzo:
Also, I'm not using `rehydrate` from `glamor` anymore because it's optional. [See here](https://github.com/threepointone/glamor/blob/master/docs/server.md)

kentcdodds:
Do we know the implications of omitting `rehydrate`?

components/layout.js
```diff
@@ -1,32 +1,34 @@
...
+const NavWrapper = glamorous.div({
+  display: 'flex',
+  justifyContent: 'space-between',
+  marginTop: 10,
+  marginRight: 20,
+  marginLeft: 20
+})
 
 export default ({children}) => {
+  css.insert(baseStyles())
```
kentcdodds: Is there any chance this will happen more than once? I'm guessing no, but want to be sure :)

__tests__/__snapshots__/docs.test.js.snap
```diff
@@ -1,140 +1,138 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`Testing Docs Renders! 1`] = `
-<div>
```
kentcdodds: We should be using https://github.com/kentcdodds/jest-glamor-react :+1:

components/layout.js
paulmolluzzo: You wouldn't use the `Layout` component more than once on a view, so I'm not sure why it'd happen multiple times. Have you seen it be called multiple times?

__tests__/__snapshots__/docs.test.js.snap
paulmolluzzo: Good call. Can update when I get to a keyboard.

paulmolluzzo:
> Do we know the implications of omitting `rehydrate`?

If you look at the [`aphrodite` README](https://github.com/Khan/aphrodite/blob/master/README.md) it says rehydration will help avoid applying duplicate styles.

> Rehydrating lets Aphrodite know which styles have already been inserted into the page. If you don't rehydrate, Aphrodite might add duplicate styles to the page.

I'm not sure if rehydration works exactly the same in `glamor`, but the link I included earlier says the API is mostly the same as `Aphrodite`, so I'm guessing that's the "hit" we take.

What do you think? Put it back in or leave it out?

components/layout.js
kentcdodds: Just wonder if this component would ever be re-rendered. I've never used next.js before so I'm not sure whether that can happen.

kentcdodds:
It doesn't seem to be hurting any by having rehydration so we may as well leave it in and avoid the potential issue.

paulmolluzzo:
If we leave in `rehydrate` we have a line that tests won't cover since it's in a `if` block looking for `window`.

components/layout.js
paulmolluzzo: Ah, when you navigate it will. Maybe I can move this line to `_document.js`. Let me investigate and I'll update accordingly.

paulmolluzzo: It's not getting reinjected, I just testing by navigating and looking at the rendered HTML.

kentcdodds: Ok, sounds good. Glamor is probably bring smart about it for us.

kentcdodds:
I'd rather do the best thing for the user than for the developer in this case ✌️

__tests__/__snapshots__/docs.test.js.snap
paulmolluzzo: @kentcdodds I added `jest-glamor-react` but the snapshots look the same. 😕

paulmolluzzo:
> I'd rather do the best thing for the user than for the developer in this case ✌️

Reverted so we're still using `rehydrate`.

__tests__/__snapshots__/docs.test.js.snap
kentcdodds: Huh, weird. It could be the renderer... Though I think it's supposed to work with that one. 😕

kentcdodds:
I'm in favor of moving this repo fast and meeting things even if they're not quite ready until this is released.

PR #18, opened 2017-05-30
Pr/prettier eslint

kentcdodds:
**What**: This adds `prettier-eslint` to the project

**Why**: So we don't worry about nit-picking over style at all. Just automatically happens like magic.

**How**: two commits, one for tooling and one for formatting. We'll want to get this merged and then deal with merge conflicts.

Feel free to ask questions :) Cool thing is that you can write your code however you like, and it'll be formatted automatically when you commit.


.eslintrc
```diff
@@ -0,0 +1,14 @@
+{
+  "extends": [
+    "kentcdodds",
+    "kentcdodds/react",
+    "kentcdodds/jsx-a11y",
+    "kentcdodds/jest",
+    "kentcdodds/prettier"
+  ],
+  "rules": {
+    "react/prop-types": "off",
```
kentcdodds: These are important, but when we want to move fast the rule is annoying.

components/footer.js
```diff
@@ -13,13 +13,18 @@ const Footer = glamorous.footer({
...
 })
 
-export default () => {
+export default PageFooter
+
+function PageFooter() {
   return (
     <Footer>
-      Made with 💙 by <Anchor external href="https://github.com/paypal">PayPal</Anchor>.
+      Made with <span role="img" aria-label="love">💙</span> by
```
kentcdodds: This is an accessibility related change.

package.json
```diff
@@ -94,8 +77,13 @@
     }
   },
   "lint-staged": {
+    ".all-contributorsrc": [
```
kentcdodds: Threw this in because it's awesome :)

kentcdodds:
I'm going to go ahead and merge this so folks can rebase.

PR #25, opened 2017-05-31
Pr/fix mobile styles

paulmolluzzo:
**What**: Fixing mobile styles on homepage and documentation

**Why**: Because it looked bad

**How**: Debugged on my phone and updated the code


paulmolluzzo:
To whoever takes a look at this PR, it should really be checked on your phone.

package-scripts.js
```diff
@@ -21,6 +21,7 @@ module.exports = {
     },
     test: {
       default: 'cross-env NODE_ENV=test jest --coverage',
+      update: 'cross-env NODE_ENV=test jest -u',
```
kentcdodds: Nice :+1:

kentcdodds:
Looking better! A few things are still kinda funny 😅 

screen shot 2017-05-31 at 12 31 29 pm

paulmolluzzo:
Images should really get a width attribute and we should make the CSS a max-width. I can fix next time at a my machine.

package-scripts.js
paulmolluzzo: This is why I prefer direct npm scripts. You can pass flags easier: `npm test -- -u`. I don't know how to pass flags with `nps`

kentcdodds: Yeah, with `nps` it would be: `nps "test -u"` or `nps "t -u"` even. I'm working on making it so you don't need the quotes so you could do: `nps t -u` :) (you'd need `nps` installed globally or the local version available in your PATH, otherwise you'd have to do: `npm start t -u`).

kentcdodds:
Let's merge this thing

paulmolluzzo:
@kentcdodds I fixed that image issue you pointed out with the huge badges.

screen shot 2017-05-31 at 11 00 50 pm

Relevant line: https://github.com/kentcdodds/glamorous-website/pull/25/commits/2b7a2c8c0ed41f168a98f31a23afc7796e508242#diff-31643f60bcbaf6adc248750b64546291R9

package-scripts.js
paulmolluzzo: I have `npm test` aliased. So I keep typing `nt -- -u` 😅

kentcdodds: I'll try to work on that feature I was talking about and then that should totally work

PR #32, opened 2017-06-01
add react-live-markdown.js

kentcdodds:
**What**: This adds a component that loads a remote markdown file and dynamically takes any `react-live` code blocks turning them into ReactLive components

**Why**: Makes it extremely easy to write docs in markdown _and_ have ReactLive examples 👌 

**How**: [Started here](https://codesandbox.io/s/xv7QBEgLz) because codesandbox is AMAZING. I didn't really change much with existing stuff. I added a new page called `/advanced` to test this stuff out. We'll want to come up with a better way to do this in the future, but I think this is going to be awesome.

Here's how things look now:

screen shot 2017-06-01 at 12 38 43 am

Definitely needs some styling work, but this is a great start!

It's actually pretty slick. Writing a plugin for remark is actually surprisingly straightforward :)


kentcdodds:
Sweet! We've now got more reliable `now` integration in our PRs. You can see the deployed version in the status checks:

screen shot 2017-06-01 at 12 35 32 pm

Click "Details" which will link you to the deployed version. Then you can go to the `/advanced` route to see this! https://kentcdodds-glamorous-website-pr-interpolation.now.sh/advanced

components/code-preview.js
```diff
@@ -27,9 +27,9 @@ const StyledLiveError = glamorous(LiveError)((props, {colors, fonts}) => ({
   padding: '1rem',
 }))
 
-const CodePreview = props => {
+const CodePreview = ({noInline = true, code}) => {
   return (
-    <StyledLiveProvider noInline code={props.code} scope={{glamorous}}>
+    <StyledLiveProvider noInline={noInline} code={code} scope={{glamorous}}>
```
paulmolluzzo: At this point we might as well pass in the `scope` as well. If we ever want an example that relies on `glamor` or some other package we'll be able to pass that in via `props`.

components/react-live-markdown.js
```diff
@@ -0,0 +1,58 @@
...
+import visit from 'unist-util-visit'
+import CodePreview from './code-preview'
+
+class ReactLiveMarkdown extends Component {
+  constructor() {
+    super()
+    this.state = {markdownComponents: <div>loading</div>}
+  }
+  componentDidMount() {
+    if (typeof window === 'undefined' || window.__NEXT_DATA__ === undefined) {
```
paulmolluzzo: are we really dependent on `window.__NEXT_DATA__` here?

components/react-live-markdown.js
```diff
@@ -0,0 +1,58 @@
...
+import axios from 'axios'
+import remark from 'remark'
+import remarkHtml from 'remark-html'
+import visit from 'unist-util-visit'
+import CodePreview from './code-preview'
+
+class ReactLiveMarkdown extends Component {
+  constructor() {
+    super()
+    this.state = {markdownComponents: <div>loading</div>}
```
paulmolluzzo: TODO: Have a loading component.

pages/advanced.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react'
```
paulmolluzzo: Do we plan to keep this page forever or is it just for example?

components/react-live-markdown.js
kentcdodds: Nope. I copy/pasted this. Still new to next :) Glad you're around! You've been awesome!

pages/advanced.js
kentcdodds: Yeah, I think we will. Seem my other comment.

kentcdodds:
> We haven't gotten rid of the client side loading and missing content.

Not sure what you mean.

> If you change the glamorous.Button in the example to glamorous.a (or some other non-component) the whole site errors out. This is really a react-live issue, but worth pointing out.

That's annoying. I wonder whether there's a way for us to catch that error 🤔 If not, maybe we can contribute back to react-live...

> Not entirely sure of the full plan here with the /advanced page, so I'd be interested to hear your thoughts on where that comes into play. I figured we'd be using something like this in /guides

Yes, this would be on `/guides`. As we're discussing in #5, we'll probably need an advanced page in addition to guides/integrations/etc.

paulmolluzzo:
> Not sure what you mean.

I mean that the page loads and then the markdown content loads. It's not being rendered on the server. I don't have a solution though. 😕

components/react-live-markdown.js
paulmolluzzo: Thanks! Happy to be a part of this!

kentcdodds:
Ah yeah, so unless there's a way for us to reliably tell what language the user wants to view on the server, we wont be able to server render stuff.

That said, we _could_ make that work with a query parameter 🤔 I'm happy to do that! But maybe we should push that off to after the release.

paulmolluzzo:
> That said, we could make that work with a query parameter 🤔 I'm happy to do that! But maybe we should push that off to after the release.

We'd want to use a path or subdomain. Query parameters are not advisable for language/locale specification. Here's one link, but there's plenty of other resources back it up: https://webmasters.googleblog.com/2010/03/working-with-multi-regional-websites.html

kentcdodds:
Sounds good to me. Could you open an issue where we can discuss this further?

paulmolluzzo:
I'll open an issue, but I think the solution to both problems is best solved by moving to documentation out of markdown and into simple react components with titles, paragraphs, and code blocks. It's how `styled-components` did it and comes with a ton of pros.

kentcdodds:
I'd love that. Just as long as we can localize the content.

paulmolluzzo:
Example from `styled-components`: https://github.com/styled-components/styled-components-website/blob/master/components/api/primary/styled.js

kentcdodds:
I'm afraid that'd be pretty tough to translate and keep translated. Is your only concern with markdown that we can't server render it without adding a little more complexity?

paulmolluzzo:
> I'd love that. Just as long as we can localize the content.

I suspect it'll be about the same to pass the locale as a state to a markdown loader as it would to pass to a component or HOC that loads components of a type.

paulmolluzzo:
> I'm afraid that'd be pretty tough to translate and keep translated. Is your only concern with markdown that we can't server render it without adding a little more complexity?

Yes. I don't want to lose SSR. The site will seem really slow and it'll reflect poorly on the `glamorous` library even though it isn't the culprit.

kentcdodds:
Hmmm... So how would we do that without shipping all localization strings of every locale to the browser? As the user navigates the app after the first load, it's a SPA at that point right?

paulmolluzzo:
Hmm, I don't believe so, no. `next` is prefetching URLs, not loading it all on the first load.

As I understand it `next` would render, say, English components and then switching locale would fetch the new language (Spanish?) versions for that page. If you navigate it'll prefetch the page in Spanish, but all the other Spanish pages aren't in the client yet (or other languages for the current page).

This is in contrast to any page in any language will load without markdown and only after loading will it fetch and load the markdown. This even occurs with prefetch because the server only provides the HTML without the markdown in it.

I don't know that we'll see a performance difference between 1 or 100 languages since only one is loaded at a time from the server. With markdown we need SSR to escape the _perceived_ performance issue since the markdown is loaded on the client.

Again, this is how I understand it and @rauchg or @arunoda or someone else more familiar with next can correct any/all of my comments. 😀

kentcdodds:
If that's the case, what if we put the markdown in a JS module that just exports it as a string? That way it's easy to translate and can be server rendered!

paulmolluzzo:
That could work. Wanna try here?

kentcdodds:
Will do!

translations/en/advanced_refs.js
```diff
@@ -0,0 +1,36 @@
...
+          Submit
+        </glamorous.Button>
+      </form>
+    );
+  }
+}
+
+render(<MyComponent />)
+~~~
+`.replace(/~/g, '`')
```
kentcdodds: This is kinda tricky, but I figure it's better than having to escape the backticks everywhere...

pages/advanced.js
```diff
@@ -0,0 +1,22 @@
...
+import glamorous, {Div} from 'glamorous'
+import Logo from '../components/glamorous-logo'
+import Layout from '../components/layout'
+import markdownToReactLive from '../components/react-live-markdown'
+
+function Advanced(props) {
+  const locale = 'en'
```
kentcdodds: We'll figure out how to make this dynamic eventually :)

translations/en/advanced_refs.js
paulmolluzzo: Neat trick! Can we add a note in the README about this, please? This is almost a new syntax for markdown.

paulmolluzzo:
I'm pretty satisfied here. Excellent work! 👏 

If you want to add one simple example to /guide that would be nice to have. This is pretty much how we'll write guides I think, so might as well make an example.

Finally, this is a nitpick, but I think the directory name "translations" could be better as "documentation" and then have directories for each language inside. We can do the same for "guides" and "examples" and then it's way easier to find what you're looking for. For example, if I want to update the English documentation for overriding component styles I go to `documentation/en/overriding-component-styles.js`

translations/en/advanced_refs.js
ajwhite: As long as you don't need ~linethroughs~, clever!

paulmolluzzo: And paths relative to the user root: `~/some/path/`

kentcdodds: That's why I'm putting the `.replace` here instead of later in the abstraction, so each file can do its own thing as needed :)

ajwhite: Keeping it downstream, smart 👍

kentcdodds:
I made a few changes (and had some [wild ideas](https://github.com/kentcdodds/glamorous-website/tree/save/a-wild-idea/translations) which I pulled back on). If I could get a review on this that'd be great. Let's get this merged soon and iterate on the design :+1:

kentcdodds:
Hmmm... There's a build error (we should probably include the actual build script in `validate`):

```
~/Developer/glamorous-website (pr/interpolation)
🍰  $ npm run build

> glamorous-website@1.0.0 build /Users/kdodds/Developer/glamorous-website
> next build

> Using external babel configuration
> location: "/Users/kdodds/Developer/glamorous-website/.babelrc"
> Using "webpack" config function defined in next.config.js.
> Failed to build on /var/folders/v3/gkts3ng93tl6lm5hy6qp78pm3kfkyl/T/6d825fb1-9ef0-4d0e-aa4d-6ca51ad3432d
{ Error: bundles/pages/examples.js from UglifyJs
Unexpected token: operator (>) [bundles/pages/examples.js:16953,22]
    at /Users/kdodds/Developer/glamorous-website/node_modules/next/dist/server/build/index.js:181:21
    at /Users/kdodds/Developer/glamorous-website/node_modules/next/node_modules/webpack/lib/Compiler.js:272:15
    at Compiler.emitRecords (/Users/kdodds/Developer/glamorous-website/node_modules/next/node_modules/webpack/lib/Compiler.js:367:37)
    at /Users/kdodds/Developer/glamorous-website/node_modules/next/node_modules/webpack/lib/Compiler.js:265:12
    at /Users/kdodds/Developer/glamorous-website/node_modules/next/node_modules/webpack/lib/Compiler.js:360:11
    at next (/Users/kdodds/Developer/glamorous-website/node_modules/tapable/lib/Tapable.js:154:11)
    at Compiler.compiler.plugin (/Users/kdodds/Developer/glamorous-website/node_modules/next/node_modules/webpack/lib/performance/SizeLimitsPlugin.js:99:4)
    at Compiler.applyPluginsAsyncSeries1 (/Users/kdodds/Developer/glamorous-website/node_modules/tapable/lib/Tapable.js:158:13)
    at Compiler.afterEmit (/Users/kdodds/Developer/glamorous-website/node_modules/next/node_modules/webpack/lib/Compiler.js:357:8)
    at Compiler.<anonymous> (/Users/kdodds/Developer/glamorous-website/node_modules/next/node_modules/webpack/lib/Compiler.js:352:14)
  errors: 
   [ 'bundles/pages/examples.js from UglifyJs\nUnexpected token: operator (>) [bundles/pages/examples.js:16953,22]',
     'bundles/pages/advanced.js from UglifyJs\nUnexpected token: operator (>) [bundles/pages/advanced.js:16953,22]' ],
  warnings: [] }
```

I'm not sure what's wrong 🤔

paulmolluzzo:
@kentcdodds It's something in `interactive-markdown.js`. I can get `npm run build` working if I strip use of that module out of `/examples` and `/advanced` pages.

I have a feeling there's a babel preset/plugin you're not including. Maybe look that over? Sorry I'm not more help, I can't sit at a desk for long right now.

kentcdodds:
It's because `strip-indent` cannot be uglified because it's using arrow syntax:

```
'use strict';
module.exports = str => {
	const match = str.match(/^[ \t]*(?=\S)/gm);

	if (!match) {
		return str;
	}

	// TODO: use spread operator when targeting Node.js 6
	const indent = Math.min.apply(Math, match.map(x => x.length)); // eslint-disable-line
	const re = new RegExp(`^[ \\t]{${indent}}`, 'gm');

	return indent > 0 ? str.replace(re, '') : str;
};
```

I could probably make babel not ignore that file, but I think instead I'll just copy it into our project.

paulmolluzzo:
Ahh! You got rid of the `dev` script? Instead of `nr dev` I have to type `ns dev` and I know my muscle memory is gonna mess up every time. 😭 LOL

Stoked to have this PR merged! Thanks @kentcdodds! 🙌

kentcdodds:
Haha, no worries, this is all for you: https://github.com/kentcdodds/glamorous-website/commit/975ad7bd6c85338847fcf0c52701081362cac028

PR #43, opened 2017-06-04
chore: add an identifier file

kentcdodds:
**What**: This adds a `get-build-info.js` script

**Why**: To be used in the creation of a `static/build-info.json` file. Because it can be pretty useful when trying to understand what you're looking at for a deployment.

**How**: added the script file and a script.


kentcdodds:
Ah, the build fails with this because there's no `git` info in now deployment. Should probably use environment variables instead. Will do that later.

PR #55, opened 2017-06-08
feat(locale): add localization chooser

### LINKED ISSUE #7: Add in Internationalization and language support
In the original [CodeSandbox](https://codesandbox.io/s/59XX9B1MR) there was support for several languages, but in the initial port to `next` it was not preserved.

We'd like internationalization that supports language selection and will modify all of the UI content (nav, footer, buttons) and when #5 is complete, should also be used to load the correct translation of the documentation.

The CodeSandbox used [`react-intl`](https://github.com/yahoo/react-intl) so whoever takes this on may want to investigate that library and use it.

kentcdodds:
And a TON of other stuff 😅

**What**: This adds the localization chooser, and a bunch of other stuff.

**Why**: co-locating our content makes maintaining things easier and I think we should be able to still use crowdin too... 🤔 

**How**:
- Move `translations` to `content` folders that are co-located to the components/pages they're for.
- Modify `locale.js` to handle this case and add more abstraction to make it easier to consume.
- Add `locale-chooser.js`
- Add a little bit of documentation (will add more later)
- Update snapshots (they're probably way too big, we should work on making them smaller because at this point they're not really useful)

Closes #7


kentcdodds:
Rats, need to fix something :-(

kentcdodds:
Don't merge this yet. I just had a great idea... But it'll have to wait until tomorrow.

kentcdodds:
Unfortunately I don't think that `now` allows you to use subdomains as aliases unless you're using a custom domain, so you can't test this out with [the staging](https://glamorous-website-mxfstqkycn.now.sh/) deploy, but it works and it's awesome. Excited to get this merged so we can try it out.

But I'd like to get a review from someone to make sure that what I'm doing isn't a really bad idea.

kentcdodds:
Oh, and I confirmed that the way I've got the content structured, we will be able to use crowdin which will be very helpful!

paulmolluzzo:
@kentcdodds sorry, can you also add the instructions for setting up `/etc/hosts` to test locally? Keeping in mind they need to support `en.localhost.dev` and `localhost.dev` (and `es` and `fr`). Thanks!

kentcdodds:
That's actually been added in the CONTRIBUTING.md file 🎉

Issue #23, opened 2017-05-30
Check accessibility


kentcdodds:
In particular, we should verify color contrast and font sizes are accessible :+1: we already have `eslint-plugin-jsx-a11y` which is really helpful for many things. Would be great to just make sure our design is accessible to people :+1:

mwickett:
I've opened #66 to address a11y issues. 

Largest issue was related to color - which means that the design changes, which is always a challenge. Definitely not trying to step on toes here, but I made some choices on modified color that meeting AA requirements, but don't tear apart the design (which looks fantastic!)

This is my second (I think) OSS PR, and your fantastic level of structure and process is new to me, so if I've missed something, let me know and I'll update.

Look forward to discussing!

chrisdemars:
Are there any other a11y issues. I would be happy to test and help make improvements where I can.

kentcdodds:
Awesome! Thanks! You can look at what we've got now at https://rc.glamorous.rocks

Or you can look at the current pull request and see the link to the staging URL (I'm the github statuses) to see what that looks like 👍

chrisdemars:
Thanks Kent. Happy to help wherever.

mwickett:
Three outstanding issues that I'm aware of:

1. The 'home' (`href="/"`) link that is wrapped around the lipstick icon in the nav needs some text (that should be hidden visually)
2. The example code on the front page - I *think* the syntax highlighting is Prism, but a lot of those colors aren't compliant, and need to be updated, but I can't find the style code for that.

kentcdodds:
Maybe @philpl can help us with the example code thing that's using react-live. Any way to change the color scheme to be more accessible?

philpl:
@mwickett The default for react-live is really minimal and admittedly bad 😆 

You can disable the default stylesheet on the provider with the `mountStylesheet` prop set to `false`. https://github.com/FormidableLabs/react-live#liveprovider- 

Then you should add this CSS code for `.prism-code` somewhere else – preferably with glamorous, I guess: : https://github.com/FormidableLabs/react-live#liveprovider- 

And after you can add some custom themes from here: https://github.com/PrismJS/prism-themes

:tada: (btw Atom Dark is nice)

mwickett:
Thanks @philpl - I'll dig into this later today.

Just FYI for others, I've updated my PR ( #66 ) to include some small color tweaks and a fix for the non-compliant 💄 home link in the nav.

chrisdemars:
[...]

PR #66, opened 2017-06-11
fix($browser): Darken primary color plus small a11y tweaks

mwickett:
Darkened the primary colors, added aria-label to locale selector and created unique IDs for logo SVG
paths

**What**: Primary colors have been changed for accessibility (AA) compliance. Also added unique ID attributes to Glamorous logo SVG path elements and added an `aria-label` attribute to the locale selector.

**Why**: Based on #23 I ran the site through a number of `a11y` checkers (primarily using [aXe](https://chrome.google.com/webstore/detail/axe/lhdoppojpmngadmnindnejefpokejbdd)) and the color contrast ratios weren't high enough.

**How**:  I darkened the primary and primaryMed color while slightly lightening the primaryLight color.

screenshot 2017-06-10 21 04 45

Additional issues were:

1. The locale selector was unlabelled. I added an `aria-label`
2. The Glamorous logo SVG paths all had `id="Shape"` which isn't AA compliant, so I created unique, description IDs

I fully realize that the color changes do alter the design - I've tried my best to maintain the same feel while acheiving AA compliance. 

There are two outstanding issues:

1. The small lipstick link back to `/` needs text (which can be hidden) - haven't been able to dig into that link styling deep enough yet. Also, I've never used 💄 yet :)
2. The home page code example section has contrast issues - I couldn't figure out where the Prism style settings are configured - point me there and I'll take a crack.


kentcdodds:
This is fantastic! Thank you very much for the contributions! I'll give this a closer look later, but I wanted to just say that this is great!

cc @jonmajorc

paulmolluzzo:
Restarting the Travis build. It failed on the deployment, not the actual tests. Those pass locally for me.

paulmolluzzo:
@kentcdodds this is the relevant error and is not really a reflection on the quality of this PR I think.

```
[deploy]   throw new Error('Missing required environment variable GH_TOKEN')
[deploy]   ^
[deploy] 
[deploy] Error: Missing required environment variable GH_TOKEN
[deploy]     at Object.<anonymous> (/home/travis/build/kentcdodds/glamorous-website/other/now-travis:16:9)
[deploy]     at Module._compile (module.js:569:30)
[deploy]     at Object.Module._extensions..js (module.js:580:10)
[deploy]     at Module.load (module.js:503:32)
[deploy]     at tryModuleLoad (module.js:466:12)
[deploy]     at Function.Module._load (module.js:458:3)
[deploy]     at Function.Module.runMain (module.js:605:10)
[deploy]     at startup (bootstrap_node.js:158:16)
[deploy]     at bootstrap_node.js:575:3
```

examples/home-page-example.js
```diff
@@ -15,13 +15,13 @@ export default locale => {
     })
 
     const solidColors = {
-      backgroundColor: '#ED5C70',
+      backgroundColor: '#DA233C',
```
paulmolluzzo: I kind of prefer `#CC3A4B` here. Is that still acceptable from an a11y perspective?

paulmolluzzo: This is the new `primaryMed` by the way.

mwickett: That works, good call.

mwickett:
@paulmolluzzo You mentioned the code snippet, and I agree. I couldn't figure out how to update the prism syntax highlighting colors. Happy to update if you can point me in the right direction.

mwickett:
Also, I failed to notice that the Travis build failed - hopefully not something I did incorrectly. If so, please let me know!

paulmolluzzo:
> @paulmolluzzo You mentioned the code snippet, and I agree. I couldn't figure out how to update the prism syntax highlighting colors. Happy to update if you can point me in the right direction.

I'm not referring to the react live area, I mean the install snippet below that:

screen shot 2017-06-10 at 10 49 17 pm

> Also, I failed to notice that the Travis build failed - hopefully not something I did incorrectly. If so, please let me know!

It's not from your code. 😉

mwickett:
@paulmolluzzo I definitely see what you mean. With the more saturated bg color, that grey doesn't work.

What about shifting to `colors.blue` - that would match `pre` elements across the site and the really dark blue looks pretty good. 

screenshot 2017-06-11 13 06 37

paulmolluzzo:
@mwickett that's better, yeah! Can you change that in this PR please?

mwickett:
Made that change @paulmolluzzo and also updated CTA button color to primaryMed instead of primary, for consistency. Think it looks better. Now, primary is so dark (because it has to be for text) that primaryMed is a more useful color in large quantities (bg color etc.)

Also added "Home" to the `/` 💄 icon link in the nav, and hid the text visually with a negative margin. 

I created a `Hidden` component using Glamorous (first time I've actually used it!), hope it's ok there. Thought it might be useful higher up in case it's needed elsewhere, but I'll look for guidance there.

```
// Used to hide text for a11y - ie. Home in navigation
const Hidden = glamorous.span({
  position: 'absolute',
  left: '-10000px',
  top: 'auto',
  width: '1px',
  height: '1px',
  overflow: 'hidden',
})
```

I believe the only outstanding issue now is the Prism syntax highlighting colors on the front page. Where do those live right now? I'm obviously missing something because I can't figure out where that's being set.

paulmolluzzo:
> Now, primary is so dark (because it has to be for text) that primaryMed is a more useful color in large quantities (bg color etc.)

Maybe we want to use that for the huge BG color on the home page then? 🤔

As for overriding the Prism styles, look here for more info: https://github.com/FormidableLabs/react-live/blob/master/README.md#liveeditor-

mwickett:
@paulmolluzzo It is pretty intense at that scale!

It's [already using `primaryMed`](https://github.com/kentcdodds/glamorous-website/blob/master/pages/index/index.js#L21). It could go to a custom, less intense color, but then I worry about losing the consistency.

Thanks for the link, now I understand how that's being rendered. Going to dig into it now.

kentcdodds:
Let's push forward with this. I'm going to go ahead and merge this and we can fix anything that's not quite right in future PRs. Thanks for a great review @paulmolluzzo and thank you so much @mwickett for this extremely valuable contribution! :clap:

PR #98, opened 2017-06-12
Update refs.md

sotayamashita:
Can I modify?

### DIFF
--- a/pages/advanced/content/refs.md
+++ b/pages/advanced/content/refs.md
@@ -14,7 +14,8 @@ reference.
 var x = 'normal code block...'
 ```
 
-```interactive {clickToRender: true, summary: 'An input in a form'}
+```
+interactive {clickToRender: true, summary: 'An input in a form'}
 const MyInput = glamorous.input({ borderRadius: 4, padding: '4px 8px' });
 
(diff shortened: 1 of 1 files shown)


pages/advanced/content/refs.md
```diff
@@ -14,7 +14,8 @@ reference.
 var x = 'normal code block...'
 ```
 
-```interactive {clickToRender: true, summary: 'An input in a form'}
+```
+interactive {clickToRender: true, summary: 'An input in a form'}
```
kentcdodds: Unfortunately not. The way this works is I need this to be parsed as a code block with the `lang` set to `interactive {clickToRender: true, summary: 'An input in a form'}`

We [use that](https://github.com/kentcdodds/glamorous-website/blob/master/components/interactive-markdown.js) to make an [interactive experience](https://rc.glamorous.rocks/advanced) for the user. It's really nice to be able to interleave these kinds of things inside the markdown. We also have a `callout` pragma, and I expect there to be more in the future as well.

sotayamashita: I understand what `interactive {clickToRender: true, summary: 'An input in a form'}` is and I will write method ignore the lang.

PR #100, opened 2017-06-12
fix(style): css for code was clipping things

kentcdodds:
Before:

screen shot 2017-06-12 at 9 12 06 am

After:

screen shot 2017-06-12 at 9 11 57 am

### DIFF
--- a/styles/base.js
+++ b/styles/base.js
@@ -81,10 +81,10 @@ export default () => `
   }
 
   pre code {
-    background-color: inherit;
-    padding: inherit;
-    border-radius: inherit;
-    font-size: inherit;
+    background-color: initial;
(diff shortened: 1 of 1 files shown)


styles/base.js
```diff
@@ -81,10 +81,10 @@ export default () => `
...
 
   pre code {
-    background-color: inherit;
-    padding: inherit;
-    border-radius: inherit;
-    font-size: inherit;
+    background-color: initial;
+    padding: initial;
+    border-radius: initial;
+    font-size: 1em;
```
kentcdodds: I tried putting `initial` for this and that made it really small. I could also put `inherit` and I think that'd work fine... 🤷‍♂️

Issue #94, opened 2017-06-12
Console fun?


kentcdodds:
Could we do something fun like this? https://twitter.com/tharakawj/status/868411676449796096
Just for fun if anyone wants to do it.

gillchristian:
I like this, lets pick a font :smile: 

I got all of them [here](http://patorjk.com/software/taag/#p=display&f=Alpha&t=glamorous). _Each picture has a link to the page with the font selected_.

[![ansi-shadow](https://user-images.githubusercontent.com/8309423/27024210-13346fa8-4f2c-11e7-94d9-fcb8cdaecc67.png)](http://patorjk.com/software/taag/#p=display&f=ANSI%20Shadow&t=glamorous)

[![big](https://user-images.githubusercontent.com/8309423/27024215-16d155ae-4f2c-11e7-9391-728b93d8e6c8.png)](http://patorjk.com/software/taag/#p=display&f=Big&t=glamorous)

[![bulbhead](https://user-images.githubusercontent.com/8309423/27024216-195f54ba-4f2c-11e7-9e0d-bed4e85ce780.png)](http://patorjk.com/software/taag/#p=display&f=Bulbhead&t=glamorous)

[![delta-corps](https://user-images.githubusercontent.com/8309423/27024219-1be4091a-4f2c-11e7-9c0d-8ef4593c165a.png)](http://patorjk.com/software/taag/#p=display&f=Delta%20Corps%20Priest%201&t=glamorous)

[![doom](https://user-images.githubusercontent.com/8309423/27024221-1e2da064-4f2c-11e7-8959-3a0df0fcd985.png)](http://patorjk.com/software/taag/#p=display&f=Doom&t=glamorous)

[![puffy](https://user-images.githubusercontent.com/8309423/27024230-23fe9976-4f2c-11e7-854c-dfbd126997e3.png)](http://patorjk.com/software/taag/#p=display&f=Puffy&t=glamorous)

[![serif-cap](https://user-images.githubusercontent.com/8309423/27024236-2724f794-4f2c-11e7-900a-b6c27a571506.png)](http://patorjk.com/software/taag/#p=display&f=Serifcap&t=glamorous)

[![sub-zero](https://user-images.githubusercontent.com/8309423/27024240-2ae17d58-4f2c-11e7-885c-029652fd6f78.png)](http://patorjk.com/software/taag/#p=display&f=Sub-Zero&t=glamorous)

And thats how it might look:

[...]

PR #101, opened 2017-06-12
feat: add console greeting message

gillchristian:
**What**: Add a console greeting message:

image

**Why**: No necessary but cool #94 

**How**: As [proposed](https://github.com/kentcdodds/glamorous-website/issues/94#issuecomment-307815805) the console greeting is implemented similar to Google Analytics is.

### DIFF
--- a/components/console-greet.js
+++ b/components/console-greet.js
@@ -1,0 +1,44 @@
+import React from 'react'
+import InlineScript from './inline-script'
+
+export default ConsoleGreet
+
+function ConsoleGreet() {
+  return <InlineScript fn={greet} />
+}
+
+function greet() {
+  const CONTRIBUTING =
+    'https://github.com/paypal/glamorous/blob/master/CONTRIBUTING.md'
+  const USERS = 'https://github.com/paypal/glamorous/blob/master/other/USERS.md'
+
+  const GLAMOROUS =
+    ' ██████╗ ██╗      █████╗ ███╗   ███╗ ██████╗ ██████╗  ██████╗ ██╗   ██╗███████╗\n' +
+    '██╔════╝ ██║     ██╔══██╗████╗ ████║██╔═══██╗██╔══██╗██╔═══██╗██║   ██║██╔════╝\n' +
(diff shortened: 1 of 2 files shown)


components/console-greet.js
```diff
@@ -0,0 +1,44 @@
...
+  console.log("%c💄 Welcome to 💄", "font-size: 24px; color: #444;")
+
+  console.log(`%c${GLAMOROUS}`, "color: #ED5C70;")
+
+  console.log("%c\n👉 We want you to contribute 👈", gradientFont)
+  console.log(`%c${CONTRIBUTING}`, "font-size: 14px")
+
+  console.log(
+    "%c\n😎 If you are using glamorous already please add it here 😎",
+    gradientFont,
```
kentcdodds: Because this code will actually not be transpiled (see [the implementation](https://github.com/kentcdodds/glamorous-website/blob/master/components/inline-script.js)) we need to make sure that the code will run in browsers we want to support (pretty much just the latest stuff). That means no fun stuff like comma-dangle on function parameter lists and calls (like this). Sad day.

I'm pretty sure that prettier is responsible for adding this. There are a few ways to avoid this. One of them is to assign things to variables so the function call lines are shorter.

On second thought... Maybe it does get transpiled 🤔 

If it does get transpiled, is there any reason you've got the `eslint-disable` lines?

gillchristian: > One of them is to assign things to variables so the function call lines are shorter.

I'll do that :+1: 

> If it does get transpiled, is there any reason you've got the `eslint-disable` lines?

I was getting the eslint error on my editor. Did not try to run the linter by CLI to check.

gillchristian: BTW it does gets transpiled

image

gillchristian:
Glad to help :smile:

paulmolluzzo:
Awesome! 😎

tdeschryver:
I'm looking at the code base to learn a thing or two and saw this component isn't using the theme colors. 
Thus it is using older colors, is this the desired behavior? If not I can create a PR :)

gillchristian:
@tdeschryver I tried to use the theme colors but the gradients did not look really well.

PR #107, opened 2017-06-12
advanced -> size

Pat-Ayres:
closes issue #75 by migrating the Size documentation from the existing docs to the new location

**What**: Added Size documentation per issue 75

**Why**: Migrating docs from existing location into the new site

**How**: docs were copied from the existing location into the new location


kentcdodds:
Thank you so much @Pat-Ayres!

PR #116, opened 2017-06-13
Syntax highlighting

### LINKED ISSUE #113: Syntax highlighting of code blocks
**Problem Description:**

Code blocks found on pages other than the homepage contain no language specific syntax highlighting.

image

**Suggested Solution:**
Provide support for syntax highlighting for shell/bash and JavaScript code blocks. The homepage appears to use Prism for its syntax highlighting (the react-live package), so it may make sense to utilize the same highlighter for other code blocks.

andrew-codes:
**What**:
Fixes #113. This PR adds support for syntax highlighting of additional pragmas (js, bash, html).

**Why**:
Static code blocks; such as `npm install glamorous` have no syntax highlighting.

**How**:
- Include Prismjs (react-live component seems to utilize this, so this was chosen for consistency)
- `utils/highlighter.js` decouples highlighting of code blocks from Prismjs and the underlying components that will render the highlighted code
    - functional approach: create a Prismjs specific highlighter, then Prismjs language specific highlighters, and use each of these for each of the new pragmas
- update basics page for bash and html code block pragmas

This is my first PR into an OSS project! If there is any feedback you are able to give on direction, code quality, or really anything, it would be greatly appreciated.

- test coverage is 100% for newly added functionality
-unfortunately, bash Prismjs support is not shipped with Prismjs and therefore requires an additional stylesheet. I do not particularly like including the stylesheet, but unsure of other options.

### DIFF
--- a/components/interactive-markdown.js
+++ b/components/interactive-markdown.js
@@ -1,10 +1,15 @@
+import Prism from 'prismjs'
+// eslint-disable-next-line import/no-unassigned-import
+import 'prismjs/components/prism-bash'
 import React from 'react'
 import remark from 'remark'
 import remarkHtml from 'remark-html'
 import visit from 'unist-util-visit'
+import StaticCodeBlock from './static-code-block'
 import CodePreview from './code-preview'
 import Callout from './callout'
 import ClickToRender from './click-to-render'
+import highlighter from './utils/highlighter'
 import stripIndent from './utils/strip-indent'
 
 export default interactiveMarkdown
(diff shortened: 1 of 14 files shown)


components/interactive-markdown.js
```diff
@@ -23,18 +28,49 @@ function ClickToRenderCodePreview(props) {
...
-      return {component, code: value, ...options}
+      const interativeCodeBlock = options.clickToRender ?
+        clickToRenderCodePreview :
+        codePreview
+      return interativeCodeBlock(options, value)
     },
     callout(options, value) {
       return {component: Callout, children: value, ...options}
     },
+    js(options, value) {
```
kentcdodds: I wonder if we could do this more generically and say: "If the language doesn't have a specific pragma handler, then use this generic one that calls `prismHighlighter(lang)` or something. Thoughts?

kentcdodds:
One unfortunate thing about using the pragma handlers and stuff is that it breaks out of the context of the markdown which busts out of the current context of HTML. Which means that if you nest a pragma inside HTML, you'll basically close the tags you're currently in. Ultimately this means you can't put a pragma inside of `<details>` which is unfortunate:

screen shot 2017-06-13 at 9 27 09 am

This is why there are `clickToRender` and `summary` options for the interactive pragma. Perhaps we could have the same thing for the others?

andrew-codes:
I was trying to keep the pattern for `ClickToRender` and `CodePreview` portions. I think I understand what you mean. Is this in the ball park of what you were thinking?

```
~~~js {summary: 'This is a summary'}
// some js code
~~~
```

I think it makes sense to do this; especially since the interactive and callout ones support this type of functionality. Assuming I understand correctly, I will update the PR with that functionality. Thanks!

kentcdodds:
I'm thinking that you'd have to opt-into it more though, so:

```
~~~js
// some js code
~~~
```

That would render as we do now, and

```
~~~js {clickToRender: true, summary: 'This is a summary'}
// some js code
~~~
```

That would render in the clickToRender thing :)

components/interactive-markdown.js
andrew-codes: It looks like the current flow is "these are the pragmas we support (interactive, callout, js, bash, html, etc.)." If I use a new pragma in the markdown content, such as `css`, I do not know or expect to have to "enable" this pragma at this LOC. It seems ideal to take unknown pragmas and try to find a highlighter for it and in the case there truly is not one, then render a basic, non-highlighted code block.

The one caveat is that Prismjs (and I suspect many other syntax highlighting libraries) have non-core languages as separate CSS files/bundles. Even though we can process `css` for example, the styles to actually render it would still be missing. In either case, I think this would be an improvement, though; add styles for new pragma versus that PLUS registering it. Does that make sense/along the lines of what you were thinking?

kentcdodds: Yep! Makes sense 👍

andrew-codes: Great. I'll close the PR until it is ready, so you don't a commit by commit play by play. ;-)

andrew-codes:
I opened a new PR #125 that is the successor to this thread. In hindsight, I should have simply left this PR open, so not to split up the conversation around it.

Issue #115, opened 2017-06-13
Suggestion: Example code block on Homepage is too big/long


wcastand:
**Problem Description:**

I think the code block in the homepage is way to big. Lot of unecessary stuff for glamorous.

When i see it, it feels like glamorous is more complex that it actually is and involve a lot of verbosity (which isn't).

You example is far too long and can't hold in my macbook's screen without scrolling :/

Overall, the site feels too big (fonts, logo, ...)

**Suggested Solution:**

I really like styled-component code block.
It's short, efficient, clear and can hold in my macbook screen without scrolling.

I think if we only style one button, it could make the homepage less "big" and more appealing to new comers.

```javascript
const Button = glamorous.a({
  border: '1px solid #F67982',
  textDecoration: 'none',
  borderRadius: 4,
  transition: 'all .3s',
  ':hover': {
    backgroundColor: ({primary}) => primary ? primary : 'white',
  }
})
```

kentcdodds:
Agreed! I've had similar feelings. Let's shorten it 👍

kentcdodds:
Thanks @wcastand!

wcastand:
i made a issue for the nav and title/button on homepage :) #121

PR #117, opened 2017-06-13
Reduce size of the code live editor and adapt size to fit in a small screen (styled-component)

wcastand:
Made the live code example smaller, reduce logo size to fit the code in a macbook 13" and delete paddingTop on code container to make the centering of element even with logo, title and button

#115


pages/index/index.js
```diff
@@ -52,6 +55,62 @@ const GettingStarted = glamorous(Button)(
...
+
+HomepageLivePreview.contextTypes = LiveContextTypes
+
+
+const StyledLivePreview = glamorous(HomepageLivePreview)({
+  padding: '1rem',
+})
+
+
+function CodePreview({noInline = true, code, scope = {glamorous}}) {
```
kentcdodds: This is better! Now we can change the other `CodePreview` without effecting this one because I think it makes sense that the `<StyledLivePreview />` is above the `<StyledLiveEditor />` on the home page, but I'd like them to be side-by-side everywhere else.

pages/index/index.js
```diff
@@ -52,6 +55,62 @@ const GettingStarted = glamorous(Button)(
...
+const StyledLiveProvider = glamorous(LiveProvider)({
+  margin: '0 auto',
+  display: 'flex',
+  flexDirection: 'column',
+  flexWrap: 'wrap',
+  justifyContent: 'space-around',
+  maxWidth: '50rem',
+})
+
+const StyledLiveEditor = glamorous(LiveEditor)({})
```
kentcdodds: If we're not going to apply styles, then I think we can just leave it as-is and just use the plain old `<LiveEditor />` below. Are we doing this in the `code-preview.js` file as well? 🙃

components/hero.js
```diff
@@ -7,13 +7,13 @@ const Title = glamorous.h1((props, {colors, mediaQueries}) => ({
   color: colors.primaryMed,
```
wcastand: I reduce the size of the title, to fit the code editor without scrolling on a 13'. the main reason is to let the visitor see the whole code instantly.
I think the live editor is the "sexy" part of the homepage, so it should be see quickly and accessible too.

I reduce the logo but i didn't want to reduce him more so i reduce the title accordingly to keep the proportions and let the logo be the biggest thing in the <Hero />.

pages/index/index.js
wcastand: in master, you do :)

kentcdodds: Whoops. Haha. Well, it doesn't really matter either way I guess.

components/hero.js
kentcdodds: Cool, I look forward to seeing what it looks like in the deployed version (hopefully that doesn't fail again)

kentcdodds:
Deploy failed again :-(

Could you take some screenshots for us to see what it looks like? Thanks!

wcastand:
Ok, so i forgot i used a bigger screen and not my 13' :/. So it's still to big for a 13"
the menu feels weird at this place thought (centered or all the way to the left makes more sense to me) :/

For me, it's still way too big overall. This is the screenshot from a 13" : 
capture d ecran 2017-06-13 a 17 48 39

and one on a bigger screen : 
capture d ecran 2017-06-13 a 17 49 55

kentcdodds:
Let's go ahead and merge this, then we can work on moving the nav and maybe even moving the buttons to one side or something to make more room above the fold :+1: Could you open an issue to discuss that?

PR #120, opened 2017-06-13
fix homepage preview code

kentcdodds:
**What**: Fixes linting and other issues on the homepage example

**Why**: Builds were breaking and content was not translated

**How**: Made `home-page-example` not a function because it doesn't have content anymore. Passed content from the index page.


kentcdodds:
screen shot 2017-06-13 at 12 17 21 pm

Issue #118, opened 2017-06-13
Improve fallback for translations


kentcdodds:
Right now we only fallback to english if the translated file doesn't exist. Instead, we should merge the english version with the translated version just in case the translated version exists, but has empty strings for some values (is incomplete) or doesn't have some values at all. This fallback behavior happens [here](https://github.com/kentcdodds/glamorous-website/blob/4bc565a4ad3257d359c99f4cc8469b82d5cf38bc/components/locale.js#L65-L87).

tdeschryver:
Is it OK if I'm taking this one?

kentcdodds:
Yes please! Thanks!

tdeschryver:
Just one quick question, does it make sense to 'replace' empty strings? In my opinion if its left blank, its done on purpose.

kentcdodds:
I think that it does. In some cases it makes sense to have an empty string for localization, but not in this docs site. I can't think of a scenario I wouldn't want a string to appear in one locale. If it's really important, people can just set it to `null` and we could do a `null` check when doing the merge maybe...

kentcdodds:
Thanks @tdeschryver!

PR #122, opened 2017-06-13
Translation fallback

tdeschryver:
Enhancement for #118.


kentcdodds:
Super! I'll review this more closely later. Do you think you could add some unit tests?

tdeschryver:
Yea sure, I was thinking about adding them but I couldn't find tests that were handling the translations so I left as it was. Ill try to add them today :),  tomorrow at the latest.

kentcdodds:
It might be a bit tricky. Thanks for giving it a try!

components/locale.js
```diff
@@ -61,14 +61,26 @@ function getInitialLocaleProps({req} = {}) {
...
 }
 
-// eslint-disable-next-line complexity
 function getContent(locale, options) {
+  const fallbackContent = content('', options)
+  if (locale === fallbackLocale) {
+    return fallbackContent
+  }
+  const localeContent = content(`${locale}/`, options)
+  return Object.keys(fallbackContent).reduce((cont, key) => {
```
kentcdodds: I'm thinking that we'll want to do a deep merge here to handle Arrays ([like here](https://github.com/kentcdodds/glamorous-website/blob/83a8d767a80dbd80724e573014bcc5c614949956/pages/basics/content/index.js#L6-L13)).

components/locale.js
```diff
@@ -79,10 +91,10 @@ function getContent(locale, options) {
       throw new Error('page or component required to get content')
```
kentcdodds: I just noticed this, could you add "example" to this error message?

components/locale.js
tdeschryver: Good that you point this one out! Completely overlooked this one

tdeschryver:
@kentcdodds thought on the testing approach? At first I tried mocking the path with `jest.mock` but without any success...

components/locale.js
tdeschryver: This might be tricky tho.

For example if the default translation is
```
module.exports = {
  sections: [
    require('./foo'), 
    require('./bar'), 
  ]
}
```
and in a translate file it is defined as 
```
module.exports = {
  sections: [
    require('./bar'), 
    require('./foo'), 
  ].filter(mod => Boolean(mod.title)),
}
```
Is there a way to link up `foo` with `foo` and `bar` with `bar`?

kentcdodds: Oh snap. You're right. Hmmm... That's a real bummer. Hmmm... Let's just do shallow for now. We'll have to circle back once things have settled a bit and maybe change how we do things a bit...

components/locale.js
```diff
@@ -61,28 +61,47 @@ function getInitialLocaleProps({req} = {}) {
...
+// eslint-disable-next-line complexity
+function content(localePath, options) {
+  const {
+    page,
+    pagePath = (path, p) => `../pages/${p}/content/${path}index.js`,
+    component,
+    componentPath = (path, c) => `../components/content/${path}${c}`,
+    example,
+    examplePath = (path, e) => `../examples/content/${path}${e}`,
+  } = options
```
kentcdodds: Nice. What if we did this instead:

```
const {
  page,
  component,
  example,
  contentPath = getContentPath()
} = options

function getContentPath() {
  if (options.page) {
    return 'blah'
  } else if (options.component) {
    return 'blah'
  } etc...
}
```

That way in the tests you could just specify `contentPath`. Thoughts?

tdeschryver: That's indeed a bit neater. You can expect a PR tomorrow.

components/locale.js
kentcdodds: Actually, you know what, we require the filename to be provided for these files, so I think we could associate it that way!

tdeschryver: Yep you're right, i'll give it a go!

tdeschryver: Can I make the assumption that `sections` will also be available in the translation pages?

kentcdodds: You can make that assumption for `pages` yes :+1:

kentcdodds: Actually, not all pages. Just some of them. Probably better do an undefined check first.

tdeschryver: When you have some time can you review the last commit please? Because I'm not 100% sure if that is the way to do it properly.

components/locale.js
```diff
@@ -61,28 +62,72 @@ function getInitialLocaleProps({req} = {}) {
...
-    if (locale === fallbackLocale) {
+    if (localePath === '') {
       throw error
     }
-    return getContent(fallbackLocale, options)
+    return {}
+  }
+}
+
+function getContentDictionary(localePath, options) {
```
kentcdodds: Maybe this should be called `requireContentFile`?

components/locale.js
```diff
@@ -61,28 +62,72 @@ function getInitialLocaleProps({req} = {}) {
...
+      return cont
     }
+
+    cont[key] = localeValue === null ? null : localeValue || fallbackValue
+    return cont
+  }, {})
+}
+function content(localePath, options) {
+  const {
+    contentDictionary = (p, opts) => getContentDictionary(p, opts),
```
kentcdodds: Maybe this should be called `requireContentFile`?

components/locale.js
```diff
@@ -61,28 +62,72 @@ function getInitialLocaleProps({req} = {}) {
...
+    return fallbackContent
+  }
+
+  return Object.keys(fallbackContent).reduce((cont, key) => {
+    const fallbackValue = fallbackContent[key]
+    const localeValue = localeContent[key]
+
+    if (key === 'sections' && fallbackValue instanceof Array) {
+      const filenames = fallbackValue.map(fbv => path.basename(fbv.filename))
+      const findFile = (sections, filename) =>
```
kentcdodds: Could we move this to the bottom of the file as a function declaration? I like doing that for small util functions.

components/locale.js
```diff
@@ -61,28 +62,72 @@ function getInitialLocaleProps({req} = {}) {
...
+  const fallbackContent = content('', options)
+  if (locale === fallbackLocale) {
+    return fallbackContent
+  }
+  const localeContent = content(`${locale}/`, options)
+  return mergeTranslations(fallbackContent, localeContent)
+}
+
+function mergeTranslations(fallbackContent, localeContent) {
+  if (!localeContent || localeContent === {}) {
```
kentcdodds: `localeContent === {}` will never be true. Maybe we could do: `Object.keys(localeContent).length === 0` :)

tdeschryver: Seems like I learned something new today! :)

components/__tests__/fixtures/content/es/foo.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  foo: 'foo in español',
+  bar: '',
+  baz: null,
+  sections: [require('./bar')],
```
kentcdodds: Maybe, to make sure the filename matching is working properly, we should make this `require('./baz')` instead.

tdeschryver: I'm not quite following, could you elaborate?
I didn't include `baz` because it should fallback to the fallback `baz`.
`bar` is included to test the nested translation.

kentcdodds: Yeah, so I just want to make sure that when the `sections` are in a different order, we use the right fallback values. So alternatively, you could take the english version and change it to: `sections: [require('./baz'), require('./bar')]` just so we're sure the fallbacks apply to the right translations :+1:

kentcdodds:
Cool! I'm deploying [here](https://zeit.co/kentcdodds/glamorous-website/mfmnokachj?redirect=1) to see what it looks like :) Thanks so much!

kentcdodds:
Tried it locally. Looks great. Thank you!

tdeschryver:
Thank you for all the inputs!

PR #125, opened 2017-06-13
Syntax highlighting

### LINKED ISSUE #113: Syntax highlighting of code blocks

andrew-codes:
**What**:
Fixes #113. This PR adds support for syntax highlighting of additional pragmas (js, bash, html).

**Why**:
Static code blocks; such as npm install glamorous have no syntax highlighting.

**How**:
Include Prismjs (react-live component seems to utilize this, so this was chosen for consistency)
utils/highlighter.js decouples highlighting of code blocks from Prismjs and the underlying components that will render the highlighted code
functional approach: create a Prismjs specific highlighter, then Prismjs language specific highlighters, and use each of these for each of the new pragmas
update basics page for bash and html code block pragmas
This is my first PR into an OSS project! If there is any feedback you are able to give on direction, code quality, or really anything, it would be greatly appreciated.

test coverage is 100% for newly added functionality
-unfortunately, bash Prismjs support is not shipped with Prismjs and therefore requires an additional stylesheet. I do not particularly like including the stylesheet, but unsure of other options.

In hindsight, I should of simply left the other PR open to be able to continue its conversation into this one. Live and learn.

### DIFF
--- a/components/interactive-markdown.js
+++ b/components/interactive-markdown.js
@@ -5,12 +5,13 @@ import React from 'react'
 import remark from 'remark'
 import remarkHtml from 'remark-html'
 import visit from 'unist-util-visit'
-import StaticCodeBlock from './static-code-block'
 import CodePreview from './code-preview'
 import Callout from './callout'
 import ClickToRender from './click-to-render'
 import highlighter from './utils/highlighter'
+import StaticCodeBlock from './static-code-block'
 import stripIndent from './utils/strip-indent'
+import {or} from './utils/specifications'
 
 export default interactiveMarkdown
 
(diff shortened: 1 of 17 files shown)


andrew-codes:
I thought I did this, but apparently I did not; I am having a difficult time updating the snapshot tests required for my changes. I have tried:

```
yarn start test -- -u # -u is not a recognized option
yarn start test -- -updateSnapshot
yarn test -- -updateSnapshot
yarn test -updateSnapshot
```

Anything I am missing?

andrew-codes:
I ended up running `NODE_ENV=test ./node_modules/.bin/jest -u` to get it to work. I am unsure if this is a user error or a legit issue to file on GitHub.

kentcdodds:
Yeah, we should probably document how to update snapshots. You can use `npm start test.update`.

I'm still concerned at the level of complexity of this code. Is there anything we can do to simplify things?

andrew-codes:
Hmm, I'm happy to take a look to simplify the code. Are there any things/area in particular that you find complicated? That should help guide me. I will markup the PR's code with areas that I think may be problem areas and, with your feedback, clean-up the ones that are problematic.

kentcdodds:
I'm just trying to follow the markdown plugin and find it difficult to follow the flow of code in my head.

components/interactive-markdown.js
```diff
@@ -23,41 +29,101 @@ function ClickToRenderCodePreview(props) {
...
-          Object.assign(codeNode, {
-            type: 'paragraph',
-            children: [{type: 'text', value: 'COMPONENT_BLOCK'}],
-            lang: null,
-            value: null,
-          })
-          return true
+        const language = getLanguage(codeNode.lang)
+        const getMatchingPragma = createSupportedPragmaMatcher()
+        const specification = or(
```
andrew-codes: These lines may be heavy-handed. I was trying to use a named function instead of an anonymous function  to increase readability and understandably of the code. The idea was to read a descriptive function name instead of parse the implementation of an anonymous function to gain understanding of what the condition is. However, with that said, the two specifications are not that complicated in their implementation, so maybe doing it the way shown is overkill.

I could see the `or` specification function being overkill also if we wanted to go more with an anonymous function here.

components/static-code-block.js
```diff
@@ -0,0 +1,26 @@
...
+        className={languageClassName}
+        dangerouslySetInnerHTML={{
+          __html: code,
+        }}
+      />
+    </pre>
+  )
+}
+
+function StaticCodeBlock({code, language, summary}) {
```
andrew-codes: This and the above component could be combined. I was thinking small and compos-able, but in reality I doubt the `CodeBlock` component will or would ever be used outside the `StaticCodeBlock`. It is not even exported and, even if it was, I wonder how much value it adds to be a separate thing.

components/utils/highlighter.js
```diff
@@ -0,0 +1,13 @@
+export default highlighter
+
+function highlighter(highlight) {
```
andrew-codes: I was trying a functional approach, but may have gone overboard. ;-) Even with the inner functions being named to, hopefully, represent what they are, the further you get into the layers, the less obvious it is what each function represents.

It leads to something like:
Create a highlighter library specific factory that can create pragma specific highlighting functionality that can handle a specific value/component combination (`code handler`). Whew, what a mouthful!

We could be less functional, reducing the number nested functions and combine some of the arguments into smaller number of functions. There may be other things to do also, what do you think?

components/interactive-markdown.js
```diff
@@ -23,41 +29,101 @@ function ClickToRenderCodePreview(props) {
...
 }
 
-function interactiveMarkdown(markdownString) {
-  const componentBlocks = []
+function highlightWithPrism(code, language) {
+  return Prism.highlight(code, Prism.languages[language])
+}
+
+function createSupportedPragmaMatcher() {
+  const prismHighlighter = highlighter(highlightWithPrism)
```
andrew-codes: Please see https://github.com/kentcdodds/glamorous-website/pull/125/files#diff-d9315b560685be6f1eba31b8e3484801R3 for more information and thoughts on how the handlers are created and setup.

components/static-code-block.js
kentcdodds: I'm fine with that. The React stuff was pretty straightforward :+1:

components/interactive-markdown.js
kentcdodds: Honestly, this is the bit that I still don't understand at all. Could you explain what this is supposed to do?

components/utils/highlighter.js
kentcdodds: Yeah, let's simplify this. Could we do a single function that takes all the args? Something like:

```
function highlighter(highlight, language, options, value) {
  const code = highlight(value, language)
  return {component, code, ...options}
}
```

And we could inline it directly in the original file?

components/interactive-markdown.js
kentcdodds: I think that I'd rather imperative code here. Even though I'm a fan of FP, I think it'd be more clear imperatively. Either that, or give the term `specification` a better name. What is that?

kentcdodds:
The explanations were helpful. Thanks for iterating on this!

kentcdodds:
I wonder whether we could really simplify things a ton if we just used [this plugin](https://github.com/ben-eb/remark-highlight.js) instead...

kentcdodds:
Hey @andrew-codes, are you still interested in doing this? I made some significant changes to content-related stuff, but merge conflicts with your branch should be fairly minimal...

kentcdodds:
Superseded by #200 
Thanks for your help on this! Gave me the right direction!

andrew-codes:
Sure thing! Sorry for not being able to get to the PR sooner than this. I'm super stoked it helped in some way, though! 👍 :-D

Issue #65, opened 2017-06-10
Add flow


kentcdodds:
That would really help I think.

kentcdodds:
Labeled this as a good first issue (with the caveat that it's only really good if you're experienced with flow already). It doesn't require you understand the existing codebase very well.

bhargav175:
Hi @kentcdodds . I would like to take this up!

paulmolluzzo:
Go for it @bhargav175! 👍

kentcdodds:
Awesome!

kentcdodds:
We've got this. Just need to add annotations as we go!

PR #132, opened 2017-06-14
Feat: integrate flow

bhargav175:
**What**:
A feature PR for jntegrating **flow-type** with the website.

**Why**:

Fix issue #65. 
To leverage static type checking with flow.

**How**:

- Added .flowconfig as well as a flow-typed directory for existing packages.
- Haven't changed any of the .js files yet. This PR only caters to handling flow integration.
- .eslintignore flow-typed directories to prevent rewrite with prettier-eslint as the flow-typed files are auto-generated.


.flowconfig
```diff
@@ -0,0 +1,12 @@
+[ignore]
+<PROJECT_ROOT>/.*/__tests__/.*
+<PROJECT_ROOT>/docs/.*
```
kentcdodds: We don't have a `/docs/` directory.

bhargav175:
@kentcdodds  I forgot to include the `flow` as npm script. When you run flow, it exits without errors. However, when you start adding the flow annotation to the js files, it works as expected, throws errors as expected. I think the config went alright.

.flowconfig
bhargav175: Oh. Right. Let me fix that.

bhargav175:
@kentcdodds Haven't actually added flow annotations to the js files yet because that would be a different feature request( which would mean a fix on potentially every js file in there). However, flow annotations do work if they are added to any of the js files. This PR only covers the setup for flow. Is that good enough ?

kentcdodds:
Yep! That's great! Thank you!

kentcdodds:
If you'd like to add flow to a file or two that would be cool!

bhargav175:
Alright. Will do that.

PR #136, opened 2017-06-14
Add flow annotations to components/utils folder

bhargav175:
**What**:
Annotate /components/utils folder with flow-type

**Why**:
To get started with flow annotations in the website.

**How**:
Added type annotations for function args, return values and variable declarations.

### DIFF
--- a/components/utils/md-to-html.js
+++ b/components/utils/md-to-html.js
@@ -4,6 +4,6 @@ import stripIndent from './strip-indent'
 
 export default mdToHTML
 
-function mdToHTML(md) {
+function mdToHTML(md: string): string {
   return remark().use(remarkHtml).processSync(stripIndent(md)).toString()
 }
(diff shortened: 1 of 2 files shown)


components/utils/md-to-html.js
```diff
@@ -4,6 +4,6 @@ import stripIndent from './strip-indent'
```
kentcdodds: Do we need `// @flow` at the top of this file?

components/utils/strip-indent.js
```diff
@@ -1,17 +1,23 @@
+//@flow
```
kentcdodds: Can we add a space here so it's `// @flow`?

components/utils/md-to-html.js
bhargav175: Yes, Looks like I forgot to add that one. Will add it.

components/utils/strip-indent.js
bhargav175: Yes. Will do.

kentcdodds:
Cool! Thanks!

PR #139, opened 2017-06-14
Lightened code element background color for a11y contrast

mwickett:
**What**:
Lightened `code` element background color.

**Why**:
To comply with WCAG 2.0 AA contrast ratio for accessibility

**How**:
Changed `background-color` value for `code` elements (in `/styles/base.js`) from #eee to #f6f6f6

### DIFF
--- a/styles/base.js
+++ b/styles/base.js
@@ -88,7 +88,7 @@ export default () => `
   }
 
   code {
-    background-color: #eee;
+    background-color: #f6f6f6;
     padding: 2px 6px;
     border-radius: 2px;
     font-size: 0.9em;


kentcdodds:
Could we get a screenshot? Thanks!

mwickett:
Good call. It's a pretty subtle change.

Original (#eee)
screenshot 2017-06-14 15 44 49

Updated (#f6f6f6)
screenshot 2017-06-14 15 40 43

The issue was really with `code a` elements as the primary colour not the grey didn't quite have enough contrast, but rather than tweaking primary colours again, or making a whole new style for links inside a code element, I thought this approach made more sense.

The monospace font is more critical to styling code than the grey background IMO.

kentcdodds:
Awesome! Thank you!

PR #142, opened 2017-06-15
fixes bad translation.

aranajhonny:
**What**: Fixes bad translation.

**Why**: 'Probar que' is not the correct translation. The correct is "Pruébalo"

**How**:  changed  'Probar que' in `pages/index/content/es/index.js`


aranajhonny:
I like the idea that the site can be read in spanish. I'll help you with what I can.

kentcdodds:
I'm glad! Thank you very much!

Issue #85, opened 2017-06-11
basics -> animation


kentcdodds:
Learn how to set up the project [here](https://github.com/kentcdodds/glamorous-website/blob/master/CONTRIBUTING.md) and how docs work [here](https://github.com/kentcdodds/glamorous-website/blob/master/other/CONTRIBUTING_DOCUMENTATION.md). For much of this you should be able to do everything without even cloning the project. You could just do it in [the GitHub editor](https://help.github.com/articles/editing-files-in-another-user-s-repository/)!

- Existing docs: The animation bit from [here](https://github.com/paypal/glamorous/blob/master/README.md#example-style-objects)
- Destination: https://github.com/kentcdodds/glamorous-website/blob/master/pages/basics/content/animation.js

Anyone can work on this **Just comment here when you're going to start**. Feel free to improve the existing docs if you're copying them over (it doesn't have to be a direct copy/paste if feel you can improve upon them).

Thanks!

Ronolibert:
Hi! I can take this one!

chrisdemars:
I was looking at this to start with it and I am not sure how to conribute to this issue. Is it just copying and pasting informaiton or is there a level of development involved?

kentcdodds:
Most of it should be copy/paste. But you can improve upon things if you like. You could make a better example if you like. Just make sure you read the linked docs 😉

kentcdodds:
And thank you!

chrisdemars:
Lol, if I could React, even JS for that matter. Copying and pasting seems the best option ha

TreTuna:
@chrisdemars Are you working on this one? Otherwise, I'll pick it up.

kentcdodds:
Why don't you go ahead @TreTuna, there are many of other opportunities :)

TreTuna:
Done. [144](https://github.com/kentcdodds/glamorous-website/pull/144)

chrisdemars:
Yeah that is fine. I have to figure out how to work in JSX lol. Sorry.

kentcdodds:
That's alright @chrisdemars! Whatever you can do is appreciated! :)

PR #144, opened 2017-06-15
docs(): add anitmation docs (basics section) #85

TreTuna:
**What**: add animations documentation from GitHub README to the Glamorous website

**Why**: importing all documentation to the Glamorous website

**How**: Copied and formatted the animation documentation from Glamorous README to the component on the Glamorous website

### DIFF
--- a/pages/basics/content/animation.js
+++ b/pages/basics/content/animation.js
@@ -1,6 +1,29 @@
 module.exports = {
   title: 'Animation',
   subtitle: '',
-  description: ``.replace(/~/g, '`'),
+  description: `
+    Try this in [your browser](https://codesandbox.io/s/31VMyP7XO)
+    ~~~js
+    // import css from glamor
+    import { css } from 'glamor'
+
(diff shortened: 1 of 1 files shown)


pages/basics/content/animation.js
```diff
@@ -1,6 +1,29 @@
 module.exports = {
   title: 'Animation',
   subtitle: '',
-  description: ``.replace(/~/g, '`'),
+  description: `
+    Try this in [your browser](https://codesandbox.io/s/31VMyP7XO)
```
kentcdodds: For codesansbox links, you can actually do [this](https://github.com/kentcdodds/glamorous-website/blob/75101c53e97f5676a9ac23d0ae9f9cc4169939a2/pages/integrations/content/pseudo.js#L14-L15) which will render like [this](https://rc.glamorous.rocks/integrations). I can go ahead and update it for you, but in the future you can do that instead :) Thanks!

Issue #129, opened 2017-06-14
Add contributors to each page


kentcdodds:
I love how [webpack's docs](https://webpack.js.org/configuration/devtool/) has avatars of those who contributed to the docs on that page. Someone wanna figure out how they do that and add that to our site?

<img width="605" alt="screen shot 2017-06-13 at 10 06 49 pm" src="https://user-images.githubusercontent.com/1500684/27114969-a373e8a6-5084-11e7-8bd6-a417396b028c.png">

tdeschryver:
Seems like they are putting the contributors in the [content files](https://github.com/webpack/webpack.js.org/blob/master/content/development/index.md) (these are markdowns). 
In the [page layout file](https://github.com/webpack/webpack.js.org/blob/master/components/page/page.jsx#L63) these contributors are being passed to a [contributor component](https://github.com/webpack/webpack.js.org/blob/master/components/contributors/contributors.jsx).

I think we can extend our content files with a contributors array and 'inject' them somewhere central, maybe in [file.js](https://github.com/kentcdodds/glamorous-website/blob/master/components/locale.js) (and of course create a contributors component)?

kentcdodds:
Perfect! Thanks for looking into it @tdeschryver! We could add that easily enough! Anyone wanna take on the challenge?

tdeschryver:
So this is a bit out of my comfort zone, but I gave it a shot.

![contributors](https://user-images.githubusercontent.com/28659384/27195488-3c8f8b3e-5207-11e7-83c0-0011df8e3d27.PNG)

kentcdodds:
Looks excellent! Let's see a pull request and we can give you some pointers 😀

paulmolluzzo:
Closed by #148

PR #148, opened 2017-06-15
Adding contributors to pages

tdeschryver:
Adding contributors to each page #129

### DIFF
--- a/components/contributors.js
+++ b/components/contributors.js
@@ -1,6 +1,7 @@
 import React from 'react'
 import glamorous, {Div} from 'glamorous'
 import {Anchor} from '../components/styled-links'
+import {withContent} from './locale'
 
 const Container = glamorous.div({
   display: 'flex',
@@ -18,7 +19,7 @@ const Img = glamorous.img({
   borderRadius: '100%',
   margin: '0 5px',
   transition: 'transform .3s',
-  ':hover': {
+  '&:focus, &:hover, &:active': {
     transform: 'scale(1.2)',
   },
 })
@@ -28,14 +29,15 @@ const contributor = username =>
     <Img src={`https://github.com/${username}.png?size=90`} />
   </Anchor>)
 
-export default function Contributors({contributorsTitle, contributors = []}) {
+export default withContent({component: 'contributors'}, Contributors)
+function Contributors({content, contributors = []}) {
   if (contributors.length === 0) {
     return <noscript />
   }
 
   return (
     <Container>
-      <h3>{contributorsTitle}</h3>
+      <h3>{content.header}</h3>
       <Div>
         {contributors.map(contributor)}
       </Div>
--- a/components/contributors.js
+++ b/components/contributors.js
@@ -17,6 +17,10 @@ const Img = glamorous.img({
   width: 45,
   borderRadius: '100%',
   margin: '0 5px',
+  transition: 'transform .3s',
+  ':hover': {
+    transform: 'scale(1.2)',
+  },
 })
 
 const contributor = username =>
--- a/components/contributors.js
+++ b/components/contributors.js
@@ -1,0 +1,40 @@
+import React from 'react'
+import glamorous, {Div} from 'glamorous'
+import {Anchor} from '../components/styled-links'
+
+const Container = glamorous.div({
+  display: 'flex',
+  flexDirection: 'column',
+  alignItems: 'center',
+  justifyContent: 'center',
+  '& > h3': {
+    margin: '20px',
+  },
+})
+
+const Img = glamorous.img({
+  height: 45,
+  width: 45,
+  borderRadius: '100%',
+  margin: '0 5px',
+})
+
+const contributor = username =>
+  (<Anchor key={username} href={`https://github.com/${username}`}>
+    <Img src={`https://github.com/${username}.png?size=90`} />
+  </Anchor>)
+
+export default function Contributors({contributors = []}) {
+  if (contributors.length === 0) {
+    return <noscript />
+  }
+
+  return (
+    <Container>
+      <h3>Contributors:</h3>
+      <Div>
+        {contributors.map(contributor)}
+      </Div>
+    </Container>
+  )
+}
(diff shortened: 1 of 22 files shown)


pages/advanced/content/fr/index.js
```diff
@@ -10,4 +10,6 @@ module.exports = {
     require('./size'),
     require('./ssr'),
   ].filter(mod => Boolean(mod.title)),
+  contributorsTitle: 'Contributeurs:',
```
kentcdodds: I'd prefer to not have this in every file. Could we find another place to do this? I have an idea if you want me to pick this up from here.

components/contributors.js
```diff
@@ -0,0 +1,44 @@
...
+    transform: 'scale(1.2)',
+  },
+})
+
+const contributor = username =>
+  (<Anchor key={username} href={`https://github.com/${username}`}>
+    <Img src={`https://github.com/${username}.png?size=90`} />
+  </Anchor>)
+
+export default function Contributors({contributorsTitle, contributors = []}) {
```
kentcdodds: If we use the `withContent` HoC on this file [like here](https://github.com/kentcdodds/glamorous-website/blob/master/components/nav.js#L166) then we could move the `contributorsTitle` localization to `components/content/contributors.js` and not have to pass it around everywhere. What do you think?

tdeschryver: I was thinking about that too! I was just figuring out how to to it properly :)

components/contributors.js
```diff
@@ -0,0 +1,44 @@
...
+  },
+})
+
+const Img = glamorous.img({
+  height: 45,
+  width: 45,
+  borderRadius: '100%',
+  margin: '0 5px',
+  transition: 'transform .3s',
+  ':hover': {
```
kentcdodds: To make this accessible, could we make it: `&:focus, &:hover, &:active`?

pages/advanced/content/fr/index.js
tdeschryver: Yea of course go for it!

kentcdodds: Actually, I _had_ an idea, and it wasn't good and was complicated. Then I thought of my other idea which I think you can totally do (even though you say you're not too familiar with React :wink:). You're doing a great job! Don't stop now! :) I think you know where to take things from here :)

kentcdodds: Let me know if you want help though :)

tdeschryver: OK, if i'm stuck ill give a shout

components/contributors.js
tdeschryver: Yep, makes sense!

tdeschryver: Just one quick question, is it possible to combine those 3 or do I have to do it this way
```
':focus': {
    transform: 'scale(1.2)',
  },
  ':hover': {
    transform: 'scale(1.2)',
  },
  ':active': {
    transform: 'scale(1.2)',
  },
```

kentcdodds: Good question! You do it like this:

```javascript
'&:focus, &:hover, &:active': {
  transform: 'scale(1.2)',
},
```

Works similar to how you'd do a normal CSS selector :)

kentcdodds:
Once you're finished, if you could get this deployed via `now` that'd be awesome. If not, I can do it for you, no trouble either way :)

components/contributors.js
tdeschryver: Ugh so was close... I tried `':focus, :hover, :active', dump mistake! Thanks for the help

components/contributors.js
tdeschryver: Got it working! :) I thought it would do some strange things (with content from the component and the content from the page). Just testing a bit more, commit is following!

components/layout.js
```diff
@@ -37,7 +38,7 @@ const Wrapper = glamorous.div((props, {fonts, colors}) => ({
   },
 }))
 
-function Layout({pathname, children, locale}) {
+function Layout({pathname, children, locale, content: {contributors}}) {
```
kentcdodds: Maybe instead of accepting this as a `content` prop, we should just accept a `contributors` prop. Then we could default it up here as well: `, contributors = []}`

pages/advanced/index.js
```diff
@@ -5,7 +5,11 @@ import PageSections from '../../components/page-sections'
 
 function Advanced({url, content, locale}) {
   return (
-    <Layout pathname={url ? url.pathname : ''} locale={locale}>
+    <Layout
+      pathname={url ? url.pathname : ''}
+      locale={locale}
+      content={content}
```
kentcdodds: With the change mentioned above you could change this to: `contributors={content.contributors}`

components/layout.js
tdeschryver: My bad forgot to clean up, this was at first like you're mentioning, changed it to content to also get the dirty `contributorsTitle`.

tdeschryver:
About the `now` deployment, can I just do that from my branch and just follow the [docs](https://zeit.co/now#features)?

kentcdodds:
Pretty much this should do it for you (assuming you're in the root directory of the project on your branch):

```
npm install -g now
now
```

Boom, done :)

It'll take a little while to deploy, but it'll give you a URL right away.

tdeschryver:
Getting errors - [log](https://zeit.co/OyZjyAEvDZFCN7OP05GQWTwT/glamorous-website/cpichqdsbp/logs)

kentcdodds:
Huh, weird. Well, just push what you've got and so long as the Travis build passes, I'll merge it :) Thanks!

kentcdodds:
This is so great! Thank you so much!

Would you like to [add yourself](../blob/master/CONTRIBUTING.md#add-yourself-as-a-contributor) to the [contributors table](../#contributors)? You could do that in another PR :)

tdeschryver:
Thanks again for your input!
I'll add myself later today/tomorrow.
Strange that `now` is crashing tho :(

PR #112, opened 2017-06-13
Translations in french [WIP]

forresst:
**What**:
Translations in french

**Why**:
Closes #110 

**How**:

Translate:
- [x] Core pages
- [x] Basis pages
- [x] Advanced pages
- [ ] Api pages
- [ ] Integrations pages
- [ ] Examples pages

PR #154, opened 2017-06-16
docs(install): Deleting an unnecessary subtitle

forresst:
**What**: Deleting an unnecessary subtitle in the installation doc

**Why**: When reviewing the documentation in French, [it was noted that the subtitle was unnecessary](https://github.com/kentcdodds/glamorous-website/pull/112#issuecomment-308445124).

**How**: Set the subtitle to empty


kentcdodds:
Thanks!

PR #162, opened 2017-06-18
docs():add api documentation note: issue with line breaks for subhead…

TreTuna:
…ings (api section) #78

**What**: add API documentation to Glamorous website

**Why**: To move one step further in all documentation on the website being filled!

**How**: Copied and formatted API documentation from GitHub README to website code

NOTE: I used h3s for each subheading in order to vary the size compared to main heading (which seems to be h2 size). However, I was unable to get line breaks to work properly for subheadings. The normal markdown style of leaving two spaces at the end of the previous line was not working. If I added direct <br>'s it worked, however, that was breaking markdown links. I was unable to find any other docs section done so far that had multiple subheadings. I'm imagining this could be an issue with the way the markdown is being interpretted. I will continue to investigate a bit.


pages/api/content/glamorous.js
```diff
@@ -1,6 +1,131 @@
 module.exports = {
-  title: '`glamorous`',
+  title: 'glamorous API',
   subtitle: '',
-  description: ``.replace(/~/g, '`'),
+  description: `
+    The ~glamorous~ function allows you to create your own ~glamorousComponentFactory~ (see 
+    [above](https://github.com/paypal/glamorous/blob/master/README.md#glamorouscomponentfactory)
```
tdeschryver: In the glamorous docs the `glamorousComponentFactory` part is indeed above the `api` part, but on this website it is underneath of it. I don't know which one is correct, but I think a change is required here.

pages/api/content/glamorous.js
```diff
@@ -1,6 +1,131 @@
...
+    ) for any component you have. For [~example~](https://codesandbox.io/s/g5kDAyB9:
```
tdeschryver: A typo in the link, it should be `[~example~](https://codesandbox.io/s/g5kDAyB9)`

pages/api/content/glamorous.js
```diff
@@ -1,6 +1,131 @@
...
+    // ~big~ is not forwarded to MyComponent because the ~rootEl~ is a ~div~ and ~big~
+    // is not a valid attribute for a ~div~ however ~id~ will be forwarded because
+    // ~id~ is a valid prop
+    ~~~
+    ### forwardProps
+
+    There are some cases where you're making a ~glamorousComponentFactory~ out of a custom component that spreads 
+    some of the properties across an underlying DOM element, but not all of them. In this case you should use 
+    ~rootEl~ to forward only the right props to be spread on the DOM element, but you can also use ~forwardProps~ 
+    to specify extra props that should be forwarded. For [example](https://codesandbox.io/s/GZEo8jOyy:
```
tdeschryver: A typo in the link, it should be `[example](https://codesandbox.io/s/GZEo8jOyy)`

paulmolluzzo: Am I crazy or is that the same link?

paulmolluzzo: Oh, I think you mean the `:` on the end. 👍

pages/api/content/glamorous.js
paulmolluzzo: I agree that we need a change, but I think having the API first is fine. So that means we just need to update this line to be `see below` so that it's accurate to the website's order.

@tdeschryver?

tdeschryver: @paulmolluzzo Good for me! 👍

pages/api/content/glamorous.js
tdeschryver: Indeed its the : at the end :)

TreTuna:
Thanks for all the comments. I should be able to update this later this evening.

kentcdodds:
Hi there! So things have changed a little bit. We're no longer using js files, we're using markdown. Please see the [updated docs](https://github.com/kentcdodds/glamorous-website/blob/master/other/CONTRIBUTING_DOCUMENTATION.md) :)

This should be much better! :tada:

kentcdodds:
Hmmmmm... Weird it says there are merge conflicts. Could you rebase your branch into master? This could help: http://kcd.im/pull-request

kentcdodds:
Thanks!

PR #165, opened 2017-06-19
Docs(french) : Add style-overrides example

forresst:
**What**: Translation in french "Add style-overrides example" (Related #157)

**Why**: Documentation in several languages is better

**How**: Translate in french


forresst:
cc @Macxim, @bulby97 and @etiennetalbot 😄

kentcdodds:
Hi there! So things have changed a little bit. We're no longer using js files, we're using markdown. Please see the [updated docs](https://github.com/kentcdodds/glamorous-website/blob/master/other/CONTRIBUTING_DOCUMENTATION.md) :)

This should be much better! :tada:

kentcdodds:
ping? Hoping to release soonish. Also #167 and #175

PR #166, opened 2017-06-20
feature(page-sections) Add links to individual page sections

paulmolluzzo:
**What**: Adds links to the headings of page sections.

**Why**: Makes it easier to share specific parts of the documentation.

**How**: Added a basic function to page-sections that converts the section title to a lowercase string with dashes.


paulmolluzzo:
@tdeschryver Would you like to take a look at this? Thanks!

Anyone else who wants to review and give a 👍  or 👎  that'd be helpful, then I can merge it in.

tdeschryver:
Looks good for me! 
Later on we can make it a bit fancy with an icon when you hover over it (like the github readme)?

paulmolluzzo:
@tdeschryver good call on the `:hover` state. I'll get that in soon.

components/page-sections.js
```diff
@@ -12,6 +12,9 @@ import mdToHTML from './utils/md-to-html'
 const repoEditRootURL =
   'https://github.com/kentcdodds/glamorous-website/edit/master'
 
+const sectionTitleAsID = title =>
+  `${title.toLowerCase().replace(/ /g, '-').replace(/`/g, '')}`
```
kentcdodds: I wonder whether it'd be safer in the long-run to use [`slugify`](https://www.npmjs.com/package/slugify) instead. Wouldn't add a TON to our bundle and would probably be nicer.

paulmolluzzo: OK, I can do that for sure. Should be able to do that tonight.

paulmolluzzo:
A few updates:
* Merged in `master`
* Added in `slugify` to create slugs instead of the fn I had before
* Added in a link svg and used that to make a nice `:hover` effect. (see below)

screen shot 2017-06-22 at 10 26 10 pm

kentcdodds:
Love it! Thanks! I think the link svg looks a little big. How about the same size ratio that we see on GitHub README links?

paulmolluzzo:
How's this change I pushed?

screen shot 2017-06-23 at 6 07 25 am

PR #169, opened 2017-06-20
Docs(spanish): add examples page.

aranajhonny:
**What**: Translation in spanish of examples page.

**Why**:  Documentation in several languages is better/

**How**: Translate in spanish


pages/examples/content/es/css-grid.js
```diff
@@ -0,0 +1,15 @@
+module.exports = {
+  title: 'CSS Grid',
+  subtitle: 'La mejor solución para el diseño CSS',
+  description: `
+    [CSS Grid Layout](https://developer.mozilla.org/fr/docs/Web/CSS/CSS_Grid_Layout)
```
tdeschryver: Since these docs are in Spanish I think it would be better to refer the Spanish mozilla docs @ [https://developer.mozilla.org/es/docs/Web/CSS/CSS_Grid_Layout](https://developer.mozilla.org/es/docs/Web/CSS/CSS_Grid_Layout)

pages/examples/content/es/css-grid.js
```diff
@@ -0,0 +1,15 @@
...
+    es una herramienta de diseño relativamente nueva para la web. Es increíblemente potente y expresiva,
+    permite nuevos diseños que antes eran muy difíciles o totalmente imposibles.
+
+    En este ejemplo, usamos la función [~@supports~](https://developer.mozilla.org/fr/docs/Web/CSS/@supports)
```
tdeschryver: Same point as above. See [https://developer.mozilla.org/es/docs/Web/CSS/@supports](https://developer.mozilla.org/es/docs/Web/CSS/@supports)

Issue #21, opened 2017-05-30
Twitter card


kentcdodds:
Before we officially announce the website, we should probably have a slick looking twitter preview card for it so when folks tweet about it, they see something awesome (in addition to the equally awesome domain: https://glamorous.rocks/)

So I'd love to get some mocks together for a good twitter card. We could do small:

<img width="607" alt="screen shot 2017-05-30 at 11 59 45 am" src="https://cloud.githubusercontent.com/assets/1500684/26597516/81c13bec-452f-11e7-9c06-f8cfa6bccf0f.png">

Or large:

<img width="607" alt="screen shot 2017-05-30 at 12 00 18 pm" src="https://cloud.githubusercontent.com/assets/1500684/26597536/90778ad8-452f-11e7-8f8a-035de2b9091d.png">

Once we have the image we want to use, we'll need to implement the twitter/facebook meta tags. I did this with JavaScript Air and this is what it looked like:

```html
<meta charset="UTF-8" />
<meta name="viewport" content="width=device-width, initial-scale=1" />
<meta name="google-site-verification" content="85n8ZBk_3hSeShlRmsVJXgDolakFG4UsMJgpy3mQyPs" />
<meta name="theme-color" content="#efdd4f" />
<meta name="author" content="JavaScript Air" />
<link rel="publisher" href="https://plus.google.com/b/105493143005968326308" />
<meta name="description" content="Episode 049 of the live JavaScript broadcast podcast. Kent C. Dodds has sunset JavaScript Air. No shows are currently scheduled." />
<meta name="description" content="Episode 049 of the live JavaScript broadcast podcast. Kent C. Dodds has sunset JavaScript Air. No shows are currently scheduled." />
<meta name="keywords" content="Maybe one day..." />
<meta name="author" content="JavaScript Air" />
<meta name="copyright" content="2016" />
<meta name="application-name" content="JavaScript Air Podcast" />
<meta property="og:title" content="Maybe one day..." />
<meta property="og:type" content="video" />
<meta property="og:image" content="https://javascriptair.com/episodes/2222-10-18/screenshot.png" />
<meta property="og:url" content="https://javascriptair.com/episodes/2222-10-18" />
<meta property="og:description" content="Episode 049 of the live JavaScript broadcast podcast. Kent C. Dodds has sunset JavaScript Air. No shows are currently scheduled." />
<meta property="og:site_name" content="JavaScript Air" />
<meta property="video:release_date" content="2222-10-18" />
[...]

PR #176, opened 2017-06-25
Twitter card

emmafallancy:
Added twitter card for meta tag and implement twitter card meta to all the pages except in _document directory

**What**: Adding twitter card implementation to pages like _Glamorous homepage_, _Basics_, _Advanced_, _Examples_, _Integration_, _API_ 

**Why**: Previously the twitter card implementation are all the same as in the homepage, throughout other pages as well. Hence, this implementation is done to accomodate for relevant pages twitter card with its relevant changes.

**How**: Created `twitter-card` function that returns meta tags and using `next/head` to populate the meta tags throughout the pages

_*Note:*_
Current look for the homepage twitter card looks the same and great as before but however, the remaining pages might require resizing of the images? If there is a need of anything to be change, let me know. Also, pardon for any sloppy and mistakes as this is my first attempt in contributing to open source 😄  The issue for this PR is #21 

Currently for `/api` twitter card, there's no description to it and by default, it takes the original description as to how the description in the homepage is.

Here is the link to the [deployment](https://glamorous-twitter-card.herokuapp.com/) and [validator for twitter card](https://cards-dev.twitter.com/validator)


components/twitter-card.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react'
+import {fallbackLocale} from './locale'
+
+function TwitterCard({
+  card = 'summary_large_image',
+  title = 'Maintainable CSS with React',
+  description = 'React component styling solved with an elegant API, small footprint, and great performance',
+  image = 'https://cloud.githubusercontent.com/assets/7799266/26611099/74c2a45c-4571-11e7-8290-dbdc822339be.png',
```
paulmolluzzo: We should move this image into the `/static` directory and serve from the actual app.

https://github.com/zeit/next.js#static-file-serving-eg-images

pages/api/index.js
```diff
@@ -10,6 +12,14 @@ function API({url, content, locale}) {
       locale={locale}
       contributors={content.contributors}
     >
+      <Head>
+        {twitterCard({
+          card: 'summary',
+          title: content.title,
```
paulmolluzzo: The `tagline` isn't the best title here, IMO. I should be more like `glamorous - ${tagline}`, so we get something like:
`glamorous - Basic Concepts`

Or we can do `Basic Concepts for glamorous`, `Basic Concepts | glamorous`, `glamorous | Basic Concepts`, etc. You get the idea.

pages/basics/index.js
```diff
@@ -10,6 +12,15 @@ function Basics({url, content, locale}) {
       locale={locale}
       contributors={content.contributors}
     >
+      <Head>
+        {twitterCard({
+          card: 'summary',
+          title: content.title,
+          description: content.heading,
```
paulmolluzzo: We get line breaks in the `description` when it renders. Not sure if that affects the way it displays on Twitter. :thinking:

emmafallancy: It came across my mind too if the line break would affect the way it displays on Twitter. So far, it looks fine based on how it's being displayed.

Here's the [twitter card `summary`](https://twitter.com/fancyfallancy/status/878929622645194752) look like. 

The other way that I could think of would be to have it's own content in the `content` folder, but not sure if that would be the way to go for now. 😅

paulmolluzzo: Ok, let's leave it alone for now?

pages/api/index.js
emmafallancy: I get get it done with `glamorous - Basic Concepts` but `Basic Concepts | glamorous` looks fine too. Whichever that you feel the best for it, I will implement it that way 😄

components/twitter-card.js
emmafallancy: 👍 I'll replace it to point to the image that we have in `/static`

pages/basics/index.js
emmafallancy: Yup, sure.

emmafallancy:
I can work on resizing of the image for `summary` cards and point the image for twitter card reference into `/static` directory.

Also, since we know that we only have 2 images/files to be used for twitter card for now, would it be fine I remove `image` parameter from the function? Was thinking to implement in this way for which image to be used:

`
image = card === 'summary_large_image' ? '/static/images/banner.png' : '/static/images/banner_small.png'
`

Otherwise, if you have any thoughts on how it should be, I am keen to know 😃 

> Is there any content we need stubbed/added for the other pages? I see we have some placeholders for now, but we'll want some content in there as well.

Apparently for now, only `API` page is not having any `description` and instead, it's using the placeholder by default. Reason being is because there's no content heading for `API` and I'm using content heading for the description.

pages/api/index.js
paulmolluzzo: I don't have a strong opinion either way, so long as the name `glamorous` is obvious every time. Maybe the first way is better.

It should match the page `title`, so whichever is chosen maybe you can update that as well?

paulmolluzzo:
@emmafallancy your method for determining the image looks good to me. I don't see a reason to make it more complex than that.

pages/api/index.js
emmafallancy: Yup, sure 👍 The `title` was implemented like `glamorous - React component styling solved 💄`. So I'll make the changes to match like in the `title`.

E.g. `glamorous - Basic Concepts`

kentcdodds:
Sweet! Could we get updated screenshots? Thanks! 😀

emmafallancy:
Hi @kentcdodds @paulmolluzzo ,
I've updated the image with the new ones. Currently, both cards will use the same image.

Here's how it looks like for the large `summary` twitter card.
Large summary twitter card
 
And here's how it looks like for the small `summary` twitter card. Looks fine for me. Any thoughts?
Small summary twitter card

components/twitter-card.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react'
+import {fallbackLocale} from './locale'
+
+function TwitterCard({
+  card = 'summary_large_image',
+  title = 'Maintainable CSS with React',
+  description = 'React component styling solved with an elegant API, small footprint, and great performance',
```
kentcdodds: Let's get this localized

emmafallancy: 👍

components/twitter-card.js
```diff
@@ -0,0 +1,36 @@
...
+  pathname,
+  locale,
+}) {
+  const urlPrefix = fallbackLocale === locale ? '' : `${locale}.`
+  const image = `https://${urlPrefix}glamorous.rocks/static/images/banner.png`
```
kentcdodds: We should probably also take into consideration the `rc` in the current alias for the site: https://rc.glamorous.rocks/

Currently, the preview does not work :-(

Perhaps we should do something with the `req` or `window.location` or something?

paulmolluzzo: This needs to work with SSR for Twitter to pick up the URLs, so I think we need to use `req`.

Another option is to use a universal config and pass that in as the base URL. So we'd get something like `https://${prefix}${BASE_URL}${pathname}`

https://github.com/zeit/next.js/tree/master/examples/with-universal-configuration

emmafallancy: @kentcdodds @paulmolluzzo Sorry about that. I can try work on this to pick up using `req`. Basically, what I will do is to have another more `req` parameter on the `twitter-card` component and pass in the `req` value from the pages.

What would you guys prefer?

emmafallancy: There's `getInitialLocaleProps` method that returns `locale` (uses `getLocaleAndHost` which return `locale` and `host`) and was wondering if using it would be acceptable to get the `BASE_URL`? Instead of returning `locale`, probably return `host` as well?

paulmolluzzo: Next has a `getInitialProps` method for each page that returns a context with the `req` in it: https://github.com/zeit/next.js/#fetching-data-and-component-lifecycle

I'm not sure about the `getInitialLocaleProps` and if it works on the server. @kentcdodds?

Also, at this point we might want to break this out into a separate issue so we can more easily discuss this than an old PR.

emmafallancy: `getInitialLocaleProps` is being used as function for `getInitialProps` in `withContent` function.

`getInitialLocaleProps` is currently being used together with `withContent` for component(s) in `pages` and I tried tweaking it in my local to return `host`. Works fine on the server to return `host`.

Also, I certainly agree with you @paulmolluzzo on breaking this into separate issue. 👍

Issue #102, opened 2017-06-12
Country flags for Localization 


jonmajorc:
What you did:
Stay relevant on Twitter... 🙃

What happened:
mentioned @kentcdodds and @paulmolluzzo and conversation happened... 😏

Suggested solution:
This cool repo we can use for [country flags](https://github.com/hjnilsson/country-flags) with localization dropdown

kentcdodds:
Could probably use some designs on this :)

jonmajorc:
My son was just born, so I won't be able to focus my time much right now 😄  

I did find in my research [Webpack](https://webpack.js.org/) does it best! I would use that has a reference 😃

paulmolluzzo:
@jonmajorc 🎉🎉🎉🎉 Congrats! Enjoy the time with your baby and get off the computer!

tdeschryver:
Maybe a stupid question, but do we really need this?
When starting on the implementation I asked myself which country flags I should use for the locales, for some inspiration I went looking at some sites which are already being mentioned here and there.
And none of these websites are doing this: [Jest](https://facebook.github.io/jest/), [webpack](https://webpack.js.org/), [styled-components (only en)](https://www.styled-components.com/docs/basics).

If its needed I guess we'll go for the most obvious flags (USA, Spain, France).

kentcdodds:
I wonder if it could be offensive to get the country wrong or something... Hmmm....

paulmolluzzo:
I don't think it would be offensive as it's fairly common practice to use the French flag for designating French and Spain for Spanish. Occasionally you do see English designated by an icon that's half U.K. and half USA, but AFAIK that's less about being politically correct and more about UX.

Keep in mind we're not designating locales, just language, so again a simple designation is generally acceptable.

tdeschryver:
I played around with this for a bit and came up with this, thoughts?
![image](https://user-images.githubusercontent.com/28659384/27507900-178a0c36-58da-11e7-8c0d-2a731f0ac1f7.png)

Or with full names:
![image](https://user-images.githubusercontent.com/28659384/27507918-6c0b7380-58da-11e7-9f25-a9f8cb9e5045.png)

paulmolluzzo:
I like the full word better. Just to be sure, "Help Translate" will be printed in the currently selected language, right?

tdeschryver:
Correct @paulmolluzzo! (it still is using the content files)

tdeschryver:
[...]

PR #177, opened 2017-06-27
WIP: resolve locale content at runtime

kentcdodds:
**What**: A bunch of stuff, but fundamentally I'm making it so we can write our content in regular markdown files.

**Why**: I noticed that we're actually shipping _everything_ under `content/` to _everyone_ due to the way webpack is bundling up our dynamic require statements.

**How**: No time to chat right now. Still a WIP. Take a look if you're interested!

A few requirements of this approach:
1. Static export builds of the website for every locale (with the `LOCALE` environment variable set)
2. Move most of the content from `.js` files to `.md` files. The babel plugin that I've written can handle both!
3. Use `import` statements (can't do `require`) and the import source must include `content` to be transpiled. These limitations could be worked on, but I think this is alright for now.

Things are busted all over, but the `/examples` page should be working :)

PR #178, opened 2017-06-27
Country flags in locale chooser

tdeschryver:
**What**: Proposition for country flags in the locale chooser

**Why**: see #102 

**How**: Made a custom select, flags are coming from [lipis/flag-icon-css](https://github.com/lipis/flag-icon-css/tree/master/flags/4x3)

**Screen:**
image

**Mobile screen:**
image


components/locale-chooser.js
```diff
@@ -2,47 +2,137 @@ import React from 'react'
...
+    const prefix = fallbackLocale === locale ? '' : `${locale}.`
+    return `${protocol}//${prefix}${host}${pathname}${search}${hash}`
+  }
+
+  return 'https://github.com/kentcdodds/glamorous-website/blob/master/other/CONTRIBUTING_DOCUMENTATION.md'
+}
+
+function getFlag(locale) {
+  const style = {width: '1em', height: '100%'}
+  switch (locale) {
```
tdeschryver: There must be a better option... I tried with `require('./svgs/${locale}.svg')` but kept bumping into problems.

components/locale-chooser.js
```diff
@@ -2,47 +2,137 @@ import React from 'react'
...
+
+  render() {
+    return (
+      <Wrapper>
+        <Toggle onClick={this.handleClick.bind(this)} isOpen={this.state.open}>
+          {localeContent(locales.find(({key}) => key === this.props.locale))}
+        </Toggle>
+        {this.state.open &&
+          <List aria-label="Locale selector">
+            {locales.map(localeItem)}
```
tdeschryver: Is it OK to use `locales` from `locale.js`, because before the locales where hard coded.

components/locale-chooser.js
```diff
@@ -2,47 +2,137 @@ import React from 'react'
...
+      </Wrapper>
+    )
+  }
 }
+export default withContent({component: 'locale-chooser'}, LocaleChooser)
 
-function changeLanguage({target: {value}}) {
-  let url
-  if (supportedLocales.includes(value)) {
+function localeToHref(locale) {
```
tdeschryver: Would be best to rewrite this function after #177 is complete?

components/locale-chooser.js
kentcdodds: I think I'm ok with this... I can't think of a better way at the moment...

components/locale-chooser.js
kentcdodds: Yes, you've got a merge conflict here now 😅 sorry!

kentcdodds:
Let me know when you've got the merge conflict resolved and I'll merge this. So excited! Thanks!

tdeschryver:
Working on it! Ill try to fix this PR today. 
Loving the changes!

components/locale-chooser.js
```diff
@@ -57,3 +130,40 @@ function getLocaleAndHost() {
...
+      return {
+        key,
+        display: 'Español',
+        flag: <EsSvg {...svgStyle} />,
+      }
+    case 'fr': {
+      return {
+        key,
+        display: 'Français',
+        flag: <FrSvg {...svgStyle} />,
```
kentcdodds: I wonder if we could improve ergonomics a little bit by doing:

```javascript
{
  Flag: require('./svgs/fr.svg'),
}
```

And then when you use the flag you do:

```javascript
(<div>
     <Flag width="1em" height="100%" /> <span>{display}</span>
   </div>)


```

tdeschryver: That's sort of what I was trying to do previously, but I ran into some errors. I'll try again, might need to add a package to load svg's if that's OK.

tdeschryver: Was trying to do `require('./svgs/${key}.svg')`

kentcdodds: Yeah, the dynamic require was probably not something that'd work here (because we're actually using a babel plugin for this) 😄

tdeschryver: Hmmm... I must be missing something.
When I'm just using the require I'm getting the error `Module parse failed: .\svgs\en.svg Unexpected token (1:0) You may need an appropriate loader to handle this file type.`
And with using `svg-react-loader` , I'm getting the error `Error: Cannot find module './svgs/en.svg'`.

I'll try again with a clear head tomorrow. Or if someone knows how to do it, help it welcome. 
It must be something really stupid tho :smile:

kentcdodds: Ah, you know what, I'll bet that the svg plugin we're using doesn't support require statements. Let's just stick with what you've got. It's not really ergonomic, but it's ok.

PR #181, opened 2017-06-30
fix broken html.

### LINKED ISSUE #114: Fix HTML generated from markdown to be valid
Got this email from Algolia:

Hi Kent,

I am having trouble creating the config for your doc.
The reason is that you have `<p>` tag inside titles like `<h1>`.
This is incorrect html, and the parser we are using (http://lxml.de/) is ignoring it completely.

Could you remove that on your side ?

kentcdodds:
**What**: removes surrounding `<p></p>` tags in `mdToHTML`

**Why**: We don't need them anywhere and we don't _want_ them in titles.

Closes #114

**How**: using the magic of `.slice`


kentcdodds:
Works!

PR #182, opened 2017-06-30
WIP: getting started guide

kentcdodds:
**What**: This:

screen shot 2017-06-30 at 3 38 27 pm

**Why**: Because it'll be a great way for people to get started with glamorous. Also, it was @tkh44's [idea](https://twitter.com/tkh44/status/875768337061617664) and he has great ideas.

**How**: I added the ability for our babel plugin to load any file as a raw string if it is in a `content` directory and includes `.raw` in the filename. This is pretty neat I think.
I also made it so we merge properties from the fallback language so if we're missing _some_ translations from any given file, those will be merged together (like the new Getting Started nav item).

I feel pretty good about merging this as-is and having someone else pick up on it to make it better :)


kentcdodds:
Check [this](https://en-eetkyzemek.now.sh/getting-started/) out :)

paulmolluzzo:
🎉😀 super cool! IDK how we'd solve this but on mobile it's unusable. 🙁

kentcdodds:
Maybe we should just not show it on small screens... `react-media` could give us a declarative way to not render links to it...

Issue #184, opened 2017-06-30
Improve a11y of the locale chooser


kentcdodds:
**Problem Description:**

If you're just using the keyboard, the locale chooser isn't very accessible. We can do better :)

![a11y](https://user-images.githubusercontent.com/1500684/27756331-d0eb6854-5db3-11e7-89d1-69bec5b43183.gif)

**Suggested Solution:**

Let's make the keyboard only experience great:
- Highlight the dropdown
- Use proper `aria` tags
- Improve the highlight of individual items

One other thing I noticed is that if you click on it, then click away, it doesn't go away and I would expect it to.

cc @tdeschryver

tdeschryver:
Closed by #185

PR #185, opened 2017-07-01
Improve a11y of the locale chooser

tdeschryver:
**What**: see #184 

**Why**: To make it accessible

**How**: 
- Highlighting: re-added outline on button, changed background to the `link` instead of the `list item`
- `Aria` tags: added `aria-labels` to `Links` and also `alt` to flag icons
- Close on click outside: used `componentDidMount` and `componentWillUnmount` to bind a `click` event to the document. Maybe there are some improvements here.

**Note**:
Also changed `getLocaleAndHost` to `getHost`, because the locale wasn't needed anymore (was a left over from #177)

a11y


components/locale-chooser.js
```diff
@@ -85,16 +88,31 @@ class LocaleChooser extends React.Component {
     open: false,
   }
 
-  handleClick() {
+  componentDidMount() {
+    document.addEventListener('click', this.clickOutside.bind(this), true)
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('click', this.clickOutside.bind(this), true)
```
lydell: `this.clickOutside.bind(this)` returns an entirely new function, so this won’t actually remove the event listener added in `componentDidMount`.

components/locale-chooser.js
```diff
@@ -85,16 +88,31 @@ class LocaleChooser extends React.Component {
     open: false,
   }
 
-  handleClick() {
+  componentDidMount() {
+    document.addEventListener('click', this.clickOutside.bind(this), true)
```
lydell: This is a bit misleading since `clickOutside` is fired even when clicking _inside_ the dropdown.

components/__tests__/__snapshots__/locale-chooser.js.snap
```diff
@@ -334,21 +329,25 @@ exports[`opened state 1`] = `
...
                   <a
-                    className="css-m4ei6j"
-                    href="about://es.blank"
+                    aria-label="Change localization to Español"
+                    className="css-1c1bzyg"
+                    href="about://es.undefinedblank"
                   >
                     <div>
                       <EsSvg
+                        alt="Español"
```
Anujan: It would be beneficial to add `lang="es"` where you use languages that don't match the `html[lang]` attribute so the screen reader knows what language to interpret it as.

components/__tests__/__snapshots__/locale-chooser.js.snap
```diff
@@ -274,21 +265,25 @@ exports[`opened state 1`] = `
...
               <li
-                className="css-l0f4u7"
+                className="css-1uz022"
               >
                 <glamorous(a)
-                  href="about://blank"
+                  aria-label="Change localization to English"
+                  href="about://undefinedblank"
                 >
                   <a
```
Anujan: Does this go `a[aria-label="Change localization to English"] > div > svg[alt="English"]`? If so, I'd recommend changing the the `alt` attribute to be blank on the SVGs here so they're considered decorative.

components/locale-chooser.js
```diff
@@ -138,19 +155,19 @@ function mapLocale(key) {
       return {
         key,
         display: 'English',
-        flag: <EnSvg {...svgStyle} />,
+        Flag: <EnSvg alt="English" {...svgStyle} />,
```
Anujan: Here too. alt attribute should be blank since the SVGs are decorative

kentcdodds:
Thanks for the review and comments @Anujan!

components/__tests__/__snapshots__/locale-chooser.js.snap
Belelros: Also add `role="presentation"` for better interoperability as stated [here](https://www.w3.org/WAI/tutorials/images/decorative/).

Belelros:
This is looking great. A few minor comments that I think could improve this even further.

Perhaps I got this wrong since the button seems a bit funky. However I'd add a label to the button that opens the selector with something like "Change language localization" or something similar.

Upon opening the selector, focus either the list of the first element (with focus) because the focus remains within the button.

Span within the list element should probably have `aria-hidden="true"` so it's not read.  The anchor already has the correct label: `Change localization to English` reading `English` would be redundant.

screenshot 2017-07-01 16 08 34

Close with ESC would be a nice touch. Return focus to button after that.

kentcdodds:
Great feedback! Much appreciated!

If you notice anything else on the site, feedback would likewise be appreciated!

https://rc.glamorous.rocks

lydell:
Navigating through the menu using the keyboard I kinda expected the menu to close when pressing Escape. ¯\\\_(ツ)\_/¯

components/__tests__/__snapshots__/locale-chooser.js.snap
karlgroves: The suggestion of adding `aria-hidden="true"` on the SVG eliminates the need for `role="presentation"`

components/__tests__/__snapshots__/locale-chooser.js.snap
karlgroves: +1 Add `lang="es"` on the SVG

components/__tests__/__snapshots__/locale-chooser.js.snap
```diff
@@ -334,21 +329,25 @@ exports[`opened state 1`] = `
...
+                    href="about://es.undefinedblank"
                   >
                     <div>
                       <EsSvg
+                        alt="Español"
                         height="100%"
                         width="1em"
                       >
                         <svg
+                          alt="Español"
```
karlgroves: Again, alt is not valid on SVG

karlgroves:
@kentcdodds can you put this somewhere for testing?

It appears that you're intending this to behave as a menu. If that's the case, you'll want to adhere to 
https://www.w3.org/TR/wai-aria-practices-1.1/#menu

Pay close attention to the keyboard interaction patterns defined.  The version of this menu that is currently on the site does not adhere to the necessary patterns.

tdeschryver:
Thanks for the feedback everyone!
@karlgroves Once I corrected all of the stuff, I'll try to spin up a testing environment with [now](https://zeit.co/now).

components/__tests__/__snapshots__/locale-chooser.js.snap
tdeschryver: Is this still needed after adding `aria-hidden="true"` on the svg?

tdeschryver:
I think I changed most of the remarks. Thanks for the feedback!
- removed `alt` tags on the flag `svg`'s
- added `aria-label` on the button
- added `lang` on `anchor`
- added `aria-hidden` on div around `svg` and display `span`, is this OK? (comments were that the SVG flag needed to have it)
- escape closes the locale picker and focuses the button
- refactored the click outside
- when opening the locale picker, the first item is being opened. Can this be reviewed please, I think there must be a better way of doing it...

@karlgroves When trying to set up the test environment the build process kept spitting out errors. I don't know why at the moment, but I'll take a look at it tomorrow if its still needed.

kentcdodds:
I added a script for building and deploying. See the CONTRIBUTING.md for instructions 👍

tdeschryver:
@kentcdodds I'm not able to build.
First had an issue with `integrations/index` (`cannot read property 'type' of undefined @ babel-generator\lib\whispace.js:50:19`).
And now I'm getting the error `Error: EEXIST, file already exists C:\...\glamorous-website\out\en\_next\0d3594bb-ff23-4d69-8c1a-9f70af64ff30\page\_document\index.js
    at fsError (C:\...\glamorous-website\node_modules\recursive-copy\lib\copy.js:377:15)
    at C:\...\glamorous-website\node_modules\recursive-copy\lib\copy.js:252:16
    at <anonymous>`. 

Tried reinstalling everything in a clean directory but without any success, any thoughts?

kentcdodds:
Hmmm... That's pretty weird. Ummm.... Yeah, I'm not sure what to try, sorry 😕

I'll try to pull your branch and deploy it myself later.

kentcdodds:
Here it is: https://en-glamorous-hhttuorgyn.now.sh

I've rebased on master as well...

I think that choosing a language and redirecting is broken.. 🤔 

Thanks!

karlgroves:
This will still need some work.

Here is the markup for a menu on the Tenon.io website:

```<!--// This LI container is part of the overall nav //-->
<li classname="dropdown">

	<button aria-haspopup="true" aria-owns="myAccountMenu" aria-label="Open Account menu" aria-expanded="true">
		<!--// Snipped other stuff, for clarity //-->
	</button>

	<ul id="myAccountMenu" role="menu" aria-label="Account Menu" tabindex="-1">

		<li role="menuitem" tabindex="-1">
			<a href="/dashboard.php#/project">Dashboard</a>
		</li>

		<li role="menuitem" tabindex="-1">
			<a href="/admin/">Admin</a>
		</li>

		<li role="menuitem" tabindex="-1">
			<a href="/editprofile.php">Profile</a>
		</li>

		<li role="menuitem" tabindex="-1">
			<a href="/messages.php">Notifications</a>
		</li>

		<li role="menuitem" tabindex="-1">
			<a href="/logout.php">Logout</a>
		</li>

	</ul>

</li>```

Note the menu/ menuitem roles, etc.

Also, take special care to read the Keyboard Interaction section of this:
https://www.w3.org/TR/wai-aria-practices-1.1/#menu

**Comment by tdeschryver** [comment]
@kentcdodds Do I take the localization of the menu bar (with the input of @karlgroves) in this PR?

**Comment by tdeschryver** [comment]
Is there a reason why the locale picker is being shown at the top of the side menu?
Because on the homepage and on the mobile menu its being shown as the last item.
If it has to be shown at the top, when how does it have to behave when being opened (over the other menu items, other menu items need to be pushed to the bottom, ....)?

 I changed it locally to see how it would look like if it was the last menu item:
image

**Comment by kentcdodds** [comment]
At first, I thought it was better at the top, but now I think it's better at the bottom so I'm good with your changes

**Comment by tdeschryver** [comment]
Going to push the changes to move it to the bottom.
Tested it quickly because I have to go. I'll test it thoroughly tomorrow.

**Comment by tdeschryver** [comment]
I think this is ready for another review.
The menu recommendations from @karlgroves deserves an own PR in my opinion.
@kentcdodds For the snapshot testing I used `mount` instead of `render` because with the `render` method the list items in the locale selector weren't in the snapshot.

File: components/locale-chooser.js [hunk 7]
```diff
@@ -51,68 +51,135 @@ const Item = glamorous.li((props, {colors, mediaQueries}) => ({
...
-            {localeItem(mapLocale('help'))}
-          </List>}
+        <List
+          id="locale-selector"
+          aria-label={content.ariaLabelList}
+          aria-hidden={!this.state.open}
+          open={this.state.open}
+          top={this.props.top}
+        >
+          {this.state.locales.map(({key, display, Flag}) =>
```
kentcdodds: Hmmm, should we have a condition to check for the `open` state?

components/__tests__/locale-chooser.js
```diff
@@ -11,48 +11,111 @@ const {supportedLocales, fallbackLocale} = require('../../config.json')
 expect.addSnapshotSerializer(serializer)
 expect.extend(matcher)
 
-test('closed state', () => {
-  const wrapper = renderLocaleChooser()
+test('it renders', () => {
+  setHost()
+  const wrapper = mountLocaleChooser()
+  expect(toJson(wrapper)).toMatchSnapshotWithGlamor()
```
kentcdodds: I think maintaining this snapshot will just be really difficult. I would rather this test be simply:

```
test('it renders', () => {
  setHost()
  mountLocaleChooser()
})
```

Or just remove it entirely.

components/locale-chooser.js
```diff
@@ -121,48 +188,57 @@ function localeToHref(locale) {
...
-  } else {
-    return {locale: fallbackLocale, host}
+  if (localePart === currentLocale) {
+    return rest.join('.')
   }
+
+  return host
 }
 
-function mapLocale(key = fallbackLocale) {
```
kentcdodds: I really dislike switch statements. They're overly verbose. Could we go back to the object map?

components/locale-chooser.js
```diff
@@ -121,48 +188,57 @@ function localeToHref(locale) {
...
-    },
+// eslint-disable-next-line complexity
+function mapLocale(key) {
+  const options = {width: '1em', height: '100%'}
+  switch (key) {
+    case 'en':
+      return {
+        key,
+        display: 'English',
+        Flag: <EnSvg {...options} />,
```
kentcdodds: Could we leave most of this stuff the way it was? I liked the way the rendering of `Flag` better the way it was before.

components/nav.js
```diff
@@ -130,11 +130,6 @@ class Nav extends React.Component {
               {content.home}
             </Hidden>
           </ListItemAnchor>
-          {top ?
```
kentcdodds: This change is :+1:

components/locale-chooser.js
tdeschryver: We sure can! 
Sorry, Must have overlooked it while rebasing.

components/locale-chooser.js
tdeschryver: To be honest I don't know.
I was thinking it would be better to hide the list with some `css`, because this way the DOM isn't exactly being re-rendered.
Based on your answer I was curious to see how others are doing it (jest and webpack), and they are doing it with `css`.

If you want to I can add a check on the `open` state.

components/__tests__/locale-chooser.js
tdeschryver: Removed it.

components/locale-chooser.js
kentcdodds: > because this way the DOM isn't exactly being re-rendered.

Why are you concerned about the DOM being re-rendered? I guess I don't really care either way... But it makes the snapshot even worse. If we remove all the snapshots then I'm :+1: with just using CSS to hide it.

PR #188, opened 2017-07-01
improve some stuff around the nav

### LINKED ISSUE #121: Move the nav menu and maybe the Title/logo/button on the homepage
**Problem Description:**

As discussed in #117 
The nav is weirdly aligned right now, it would be better centered or all the way to the right/left.
If we can make more size and simplify the homepage or reduce the size of the title/logo.

**Suggested Solution:**

Move the nav to the right/left or center it and move the button to make more size on homepage.
styled-component's website is a good example of simplicity i think.

kentcdodds:
**What**: Moves the nav to the left.

**Why**: Closes #121

**How**: code and stuff


kentcdodds:
Could still use a little work (especially on mobile) but it's a good start: https://en-rlyghioczn.now.sh/

components/layout.js
```diff
@@ -37,18 +37,25 @@ const Wrapper = glamorous.div((props, {fonts, colors}) => ({
   },
 }))
 
-function Layout({pathname, children, contributors}) {
+function Layout({pathname, children, contributors, topNav = false}) {
```
paulmolluzzo: Why do we want to default it to `false`?

components/nav.js
```diff
@@ -52,124 +54,109 @@ const ListItem = glamorous.li({
...
+      textAlign: 'center',
+      width: '100%',
+      padding: isOpen ? '1rem 0' : 0,
+      maxHeight: isOpen ? '100%' : 0,
+      opacity: isOpen ? 1 : 0,
+    },
+    [mediaQueries.mediumUp]: {
+      display: 'flex',
+      justifyContent: top ? 'center' : 'flex-start',
+      flexDirection: top ? 'row' : 'column',
```
paulmolluzzo: With this line the mobile version is breaking on any page we don't update to have `top={true}`. Since this is a new prop only index and Getting Started are correct on mobile right now.

pages/getting-started/index.js
```diff
@@ -57,7 +57,7 @@ function CodePreview() {
 
 function Page({url}) {
   return (
-    <Layout pathname={url ? url.pathname : ''}>
+    <Layout pathname={url ? url.pathname : ''} topNav={true}>
```
paulmolluzzo: Needs to be on all pages.

components/docs-page.js
```diff
@@ -25,7 +25,7 @@ const PageWrapper = glamorous.div((props, {colors}) => ({
   '& > h3': {
     width: '100%',
     margin: '20px auto',
-    maxWidth: '50rem',
+    maxWidth: '70rem',
```
paulmolluzzo: This should be a global variable or something I think... 🤔

wcastand:
Im on iPhone right now so I looked a bit more later but in iPhone I have a different color for the bottom border of the menu to the left 
I don't know if that's something you wanted or not.

On mobile the button get started is really big or the text is too long

Other than that, it looks really cool on iPhone :)
Looking forward the desktop version :)

paulmolluzzo:
> Im on iPhone right now so I looked a bit more later but in iPhone I have a different color for the bottom border of the menu to the left
I don't know if that's something you wanted or not.

@wcastand the border is a `div` with three colors left to right. That's how it was designed. 😀

paulmolluzzo:
Three more issues for me:

The default option is missing in the nav I think:
screen shot 2017-07-01 at 5 04 20 pm
screen shot 2017-07-01 at 5 04 27 pm

The colors for contributors and the sidebar is identical, so this part looks odd. Maybe we want a right border on the sidebar?
screen shot 2017-07-01 at 5 04 16 pm

components/layout.js
kentcdodds: Docs pages (most of the pages) will have it on the left.

pages/getting-started/index.js
kentcdodds: Only applies to `getting-started` and `index`. All the others are not on the top.

paulmolluzzo: Even on mobile? I would think that on mobile it would be on top on all pages.

components/layout.js
paulmolluzzo: Ok, it makes more sense after seeing the desktop version.

pages/getting-started/index.js
paulmolluzzo: This is what I'm referring to:

img_6064
img_6065

components/nav.js
paulmolluzzo: If you move the `mediumDown` media query underneath this I think it'll fix the mobile styles for all pages. The idea is that it's sidebar on desktop wherever we don't say it's a top nav, and across the top on mobile on all pages without needing a prop.

pages/getting-started/index.js
kentcdodds: Sorry, on mobile we'll use media queries I think.

components/layout.js
```diff
@@ -37,18 +37,25 @@ const Wrapper = glamorous.div((props, {fonts, colors}) => ({
...
   css.insert(baseStyles())
   return (
     <ThemeProvider theme={GlobalStyles}>
       <Wrapper>
-        <Div position="relative" zIndex={1}>
-          <Div display="flex" justifyContent="flex-end" alignItems="center">
-            <Nav pathname={pathname} />
+        <Div position="relative" zIndex={1} display={topNav ? null : 'flex'}>
```
paulmolluzzo: This needs to be responsive to media queries, so probably want to pull this out of the inline css props syntax.

components/layout.js
```diff
@@ -37,18 +37,25 @@ const Wrapper = glamorous.div((props, {fonts, colors}) => ({
...
+          <Div
+            display="flex"
+            justifyContent={topNav ? 'flex-end' : 'flex-start'}
```
paulmolluzzo: Dupe: This needs to be responsive to media queries, so probably want to pull this out of the inline css props syntax.

components/layout.js
```diff
@@ -37,18 +37,25 @@ const Wrapper = glamorous.div((props, {fonts, colors}) => ({
...
+            alignItems="center"
+            flexDirection={topNav ? 'row' : 'column'}
```
paulmolluzzo: Dupe: This needs to be responsive to media queries, so probably want to pull this out of the inline css props syntax.

components/layout.js
```diff
@@ -37,18 +37,25 @@ const Wrapper = glamorous.div((props, {fonts, colors}) => ({
...
+          >
+            <Nav pathname={pathname} top={topNav} />
+          </Div>
+          <Div flex="1">
```
paulmolluzzo: Dupe: This needs to be responsive to media queries, so probably want to pull this out of the inline css props syntax.

components/nav.js
paulmolluzzo: Also need to override the `width`, `maxHeight`, `justifyContent`, and `flexDirection` on mobile.

paulmolluzzo:
@kentcdodds I pushed some changes to respect the top/sidebar versions and make both responsive for mobile screens. If you take a look and approve then go for it! If you have improvements to make I'll review what you add in.

paulmolluzzo:
Not sure about the failing CI... 😕

kentcdodds:
My bad. I Brooke the scripts. Will fix in a moment.

kentcdodds:
This is great! I love the opacity stuff.

There are a few issues you'll probably notice pretty quickly as you resize the window. I don't have time to get that fixed in the next few days, so feel free if you have time :)

https://en-glamorous-jabjandwxo.now.sh

paulmolluzzo:
@kentcdodds I updated the nav to switch to the "mobile" layout on tablets. The overflowing nav is too much content to make look good once you get around 1000px.

I also increased the `h1` size on mobile and tablets. It was way too small.

kentcdodds:
Looking much better! But we need to fix this (shouldn't clip the right off).

screen shot 2017-07-02 at 8 32 35 am
screen shot 2017-07-02 at 8 32 40 am

paulmolluzzo:
Sorry, can you circle the issue? IDK what I'm looking for. 😅

kentcdodds:
screen shot 2017-07-02 at 8 32 40 am
screen shot 2017-07-02 at 8 32 35 am

paulmolluzzo:
It's the `flex`. Off to the pool now so I'll fix later. 😎

PR #192, opened 2017-07-03
Change example page content to markdown

tdeschryver:
**What**: Changes content from `js` to `md` for the example page
**Why**: Because right now the page isn't being translated, and also to unify the way for the content.
**How**: Creating markdown files and copy pasting the text! 
**Needs translation:**
es:
  - heading in `index`

fr:
  - heading in `index`
  - translating `style-overrides` (#165)

**Note:**
Is it OK to remove `codeSandboxId` in translation pages? There are all the same right now and if their would be a new sandbox, it has to be changed in every translation (if its not being forgotton).


kentcdodds:
Thanks!

PR #195, opened 2017-07-04
adding the first german translations

kentcdodds:
**What**: Adding German translations (via Google Translate) to getting started :)

**Why**: Because we care about Germans :)


config.json
```diff
@@ -1,4 +1,4 @@
 {
-  "supportedLocales": ["en", "es", "fr"],
+  "supportedLocales": ["en", "es", "fr", "de"],
```
kentcdodds: Woo! German!

pages/basics/content/de/getting-started.md
```diff
@@ -0,0 +1,27 @@
+---
```
kentcdodds: Caveat, I do not know German!

kentcdodds:
Check it out! :smile: https://de-glamorous-rputdehbsp.now.sh

components/__tests__/locale-chooser.js
```diff
@@ -19,28 +21,35 @@ test('closed state', () => {
...
+  const wrapper = mountLocaleChooser()
   wrapper.find('button').simulate('click')
-
   const links = wrapper.find('a')
-  expect(links.length).toBe(4)
+  expect(links.length).toBe(supportedLocales.length + 1)
   expect(links.at(0).getDOMNode().getAttribute('href')).not.toEqual(
     expect.stringContaining('en.'),
   )
-  expect(links.at(1).getDOMNode().getAttribute('href')).toEqual(
```
kentcdodds: Changed this so we don't have to update the tests every time we add a locale.

components/__tests__/locale-chooser.js
```diff
@@ -1,16 +1,18 @@
...
 import LocaleChooser from '../locale-chooser'
 
+const {supportedLocales, fallbackLocale} = require('../../config.json')
+
 expect.addSnapshotSerializer(serializer)
 expect.extend(matcher)
 
 test('closed state', () => {
-  const wrapper = localeChooser()
+  const wrapper = renderLocaleChooser()
```
kentcdodds: Using `render` instead of `mount` so the snapshot is a little cleaner.

package-scripts.js
```diff
@@ -47,12 +47,9 @@ module.exports = {
     // default is run when you run `nps` or `npm start`
     default: 'next start',
     dev: 'next',
-    build: Object.assign(
-      {
-        default: concurrent.nps(...supportedLocales.map(s => `build.${s}`)),
-      },
-      localeBuilds
-    ),
+    build: Object.assign(localeBuilds, {
```
kentcdodds: This busted things when you run: `npm start build.de` which would assume you were prefixing for `default`, which we're not, we want to build the german version.

package.json
```diff
@@ -11,11 +11,7 @@
   },
   "author": "Paul Molluzzo <paul@molluzzo.com> (https://paul.molluzzo.com)",
   "license": "MIT",
-  "keywords": [
```
kentcdodds: Autoformatting via prettier 🙄 😉 :smile:

package.json
```diff
@@ -89,15 +85,9 @@
         "statements": 50
       }
     },
-    "testPathIgnorePatterns": [
-      "/fixtures/"
-    ]
+    "testPathIgnorePatterns": ["/fixtures/"]
   },
   "lint-staged": {
-    ".all-contributorsrc": [
```
kentcdodds: This busted tons of people and isn't worth investigating what's up.

components/__tests__/locale-chooser.js
tdeschryver: I'm currently refactoring the tests at #185 and wondered if we really want to do the snapshot testing here.
Because imo  the rendering has so many lines of code that it isn't clear and it slows down the testing by some seconds.
I would suggest testing the rendering of the localization links (this test is must).
And if its needed we could also test some of the behaviors: e.g. the click event shows the list and sets the  `aria-expanded` to `true`, selected the first item on being opened, ...

kentcdodds: I'm happy to add more tests for more confidence here. Feel free to add more :)

PR #197, opened 2017-07-04
Fix size badges for es translation

tdeschryver:
**What**: The badges aren't being displayed in the es translation, [link](https://es.rc.glamorous.rocks/advanced/#Optimizando-el-tamano-de-la-distribucion) (scroll down a bit).

**Why**: Because it isn't pretty

**How**:  Using an inline link instead of a reference-style link, like in the other languages

**Note**: Maybe the reference-link needs some investigation or perhaps it would be best to add something about it in the `readme`?


kentcdodds:
Thanks!

In the future, if you make your pull request branches off of https://github.com/kentcdodds/glamorous-website.git rather than your fork, your PRs will be autodeployed :+1:

tdeschryver:
@kentcdodds OK, was out of habit :)

PR #210, opened 2017-07-09
docs: (mostly) add motivation docs

kentcdodds:
Also update some styling stuff and other random things to make things
nice. Please merge :)

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -14,5 +14,8 @@
     "import/no-unresolved": "off",
     "import/no-unassigned-import": "off",
     "eol-last": "off", // the crowdin stuff doesn't include a line ending 🙄
+  },
+  "globals": {
+    "Prism": false,
   }
 }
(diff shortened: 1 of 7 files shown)


.eslintrc
```diff
@@ -14,5 +14,8 @@
     "import/no-unresolved": "off",
     "import/no-unassigned-import": "off",
     "eol-last": "off", // the crowdin stuff doesn't include a line ending 🙄
+  },
+  "globals": {
+    "Prism": false,
```
kentcdodds: Weird things happened with highlighting if I tried to import Prism, so I'll just use the global :(

components/docs-page.js
```diff
@@ -122,7 +122,14 @@ function DocSection(props) {
           dangerouslySetInnerHTML={{__html: mdToHTMLUnwrapped(title)}}
         />
       </Anchor>
-      <h4 dangerouslySetInnerHTML={{__html: mdToHTMLUnwrapped(subtitle)}} />
+      <glamorous.Div
```
kentcdodds: It's not really semantically accurate to use an `h4` and it looks not awesome. So I formatted this a little differently in a way I think is nicer

paulmolluzzo: A `div` doesn't seem semantically accurate either. 😕 Maybe don't use `mdToHTMLUnwrapped` so we at least get a `p` tag? I could be missing something though...

kentcdodds: Oh, good idea. I'll do that.

PR #224, opened 2017-07-10
Getting started - remove scrollbar

tdeschryver:
**What**: Removing the scroll bar at the right side of getting started


kentcdodds:
Thank you!

PR #219, opened 2017-07-10
Feature/search

RIanDeLaCruz:
**What**: Added Search functionality via Algolia

**Why**: Site Usability

**How**: Using Algolia DocSearch

PR #225, opened 2017-07-10
Design adjustments

motleydev:
**What**:
I've made a series of design changes such as layout, margins and max-width settings.

**Why**:
The readability is not very good in its current state.

**How**:
Exclusively through style adjustments.

**Before**

screen shot 2017-07-10 at 20 58 05

**After**

screen shot 2017-07-10 at 20 59 48

My feelings won't be hurt if you hate my style preference! :) Thanks!

### DIFF
--- a/components/docs-page.js
+++ b/components/docs-page.js
@@ -71,12 +71,12 @@ function PageSections({title, note, heading, sections}) {
       <PageWrapper>
         <Title dangerouslySetInnerHTML={{__html: mdToHTMLUnwrapped(title)}} />
         <Div
-          maxWidth="35rem"
+          maxWidth="50rem"
           margin="auto"
           dangerouslySetInnerHTML={{__html: mdToHTMLUnwrapped(heading)}}
         />
         <Div
-          maxWidth="35rem"
+          maxWidth="50rem"
           margin="20px auto"
           borderLeft="3px solid"
           paddingLeft={20}
@@ -100,7 +100,7 @@ function DocSection(props) {
     width: '100%',
     margin: '20px auto 75px',
     paddingBottom: 20,
-    maxWidth: '35rem',
+    maxWidth: '50rem',
   }))
 
   return (
--- a/components/docs-page.js
+++ b/components/docs-page.js
@@ -31,6 +31,9 @@ const PageWrapper = glamorous.div((props, {colors}) => ({
     fill: `${colors.primary}`,
     width: '1rem',
   },
+  '& img': {
+    width: '100%',
+  },
 }))
 
 const EditAnchorWrap = glamorous(Div)({
@@ -68,12 +71,12 @@ function PageSections({title, note, heading, sections}) {
       <PageWrapper>
         <Title dangerouslySetInnerHTML={{__html: mdToHTMLUnwrapped(title)}} />
         <Div
-          maxWidth="70rem"
+          maxWidth="35rem"
           margin="auto"
           dangerouslySetInnerHTML={{__html: mdToHTMLUnwrapped(heading)}}
         />
         <Div
-          maxWidth="70rem"
+          maxWidth="35rem"
           margin="20px auto"
           borderLeft="3px solid"
           paddingLeft={20}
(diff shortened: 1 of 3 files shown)


kentcdodds:
This is great! Thank you so much for these contributions! Do you think you could follow [the instructions](https://github.com/kentcdodds/glamorous-website/blob/master/CONTRIBUTING.md#making-a-pull-request) for deploying a version of this for people to look at?

I was just thinking yesterday that the font was too thin, I think you may have made it too thick though, maybe somewhere in the middle? :wink:

styles/base.js
```diff
@@ -144,7 +141,7 @@ export default () => `
   /* to be used by our code blocks */
   .prism-code {
     display: block;
-    white-space: pre;
+    white-space: pre-wrap !important;
```
kentcdodds: If `!important` is needed here, let's add a comment explaining why.

components/docs-page.js
```diff
@@ -68,12 +71,12 @@ function PageSections({title, note, heading, sections}) {
       <PageWrapper>
         <Title dangerouslySetInnerHTML={{__html: mdToHTMLUnwrapped(title)}} />
         <Div
-          maxWidth="70rem"
+          maxWidth="35rem"
```
kentcdodds: I feel that `35rem` is a little too narrow... How about `50rem`? I think that's where we had it at one time...

motleydev:
Your wish is [my url link](https://en-glamorous-rnbcwlrzvp.now.sh/). 

* I'll wait for feedback after you've poked around the live version then I'll add the comment about stuff like the `important!`.
* Sadly I'm not sure there is a _lighter_ version of the font, that's already 400 and the next lower version in the family is 100.
* 50em still seems to break the [50-60 char/line](https://baymard.com/blog/line-length-readability) But I'm happy to oblige. :) Let me know what you think of it "live"

Do it Live

kentcdodds:
Ok, the font actually looks great :+1: thanks!

I still think it's a bit too narrow. Especially on my 4k monitor 😅 

screen shot 2017-07-10 at 2 41 22 pm

In all seriousness, it also makes the interactive code blocks pretty narrow:

screen shot 2017-07-10 at 2 41 26 pm

We take a lot of hints from styled-components.com and they're doing `56rem`. I think that `50rem` should be ok without causing any trouble for folks.

Thanks again! This is great 👍

styles/base.js
```diff
@@ -144,8 +141,10 @@ export default () => `
   /* to be used by our code blocks */
   .prism-code {
     display: block;
-    white-space: pre;
-
+    white-space: pre-wrap !important;
```
kentcdodds: These styles in particular are borrowed from react-live, which injects them again. We could probably tell react-live to not inject the styles. But this is ok for now.

kentcdodds:
Would you like to [add yourself](../blob/master/CONTRIBUTING.md#add-yourself-as-a-contributor) to the [contributors table](../#contributors)?

motleydev:
NP! Happy to help! I've learned a lot from watching your Egghead videos so I'm stoked to be able to give a little back!

kentcdodds:
Haha, @motleydev, could you fix this on [the advanced page](https://rc.glamorous.rocks/advanced/) please? 😂 

screen shot 2017-07-11 at 9 19 27 pm

kentcdodds:
Those are shields.io badges and should _not_ have a width of 100% 😄

motleydev:
Just say down to do this but looks like @tdeschryver just beat me to it with [#231](https://github.com/kentcdodds/glamorous-website/pull/231)!

tdeschryver:
Oh sorry, I didn't see this :smile:

kentcdodds:
That went pretty quick! You seem to be interested in design! Maybe you can help with this! https://github.com/kentcdodds/glamorous-website/pull/219

motleydev:
Love to! Heading in to take a look.

motleydev:
Wrapping up my day here in DE. I'm going to have to take a look at this tomorrow.

PR #229, opened 2017-07-12
add translation status to the UI

### LINKED ISSUE #218: Add metadata to translated files indicating whether they're up to date
**Problem Description:**

Some translated files are out of date. This could cause people to view them and think that's all there is when in reality there's more updated information.

**Suggested Solution:**

Let's add metadata to the content object [here](https://github.com/kentcdodds/glamorous-website/blob/83c98ee6d0f9c81230b83905f3a6dcef9f967ba1/other/babel-plugin-l10n-loader.js#L82-L89) indicating whether the translation is updated so we can use that info in the UI.

Could probably use some of the logic we use [here](https://github.com/kentcdodds/glamorous-website/blob/83c98ee6d0f9c81230b83905f3a6dcef9f967ba1/other/list-l10n.js#L23-L38)

I'm gonna say that this is a must for release. I don't want to release locales with outdated translations and I think we can probably get this done faster than we can get the translations done.

kentcdodds:
**What**: This adds a small indicator to the UI for whether the translation is missing or outdated

**Why**: So we can encourage the readers to help, also so they know what they're missing

**How**: Added the metadata via the babel plugin, added a simple component for display:

screen shot 2017-07-11 at 9 53 58 pm

screen shot 2017-07-11 at 9 54 09 pm

I just thought of a thing or two more I want to do for this, so feel free to review what is here so far. I'll be back with more.

Closes #218


kentcdodds:
https://fr-glamorous-dguwxmhofp.now.sh

PR #230, opened 2017-07-12
docs(zh): translate /component into Chinise

liadbiz:
just translate /components... @bernard-lin If you have time :smile:


bernard-lin:
Looks great to me!

kentcdodds:
cc @yuanchuan @HsinChiehYeh 😀

yuanchuan:
LGTM

components/content/zh/page-sections.md
```diff
@@ -0,0 +1,3 @@
+---
+edit: '编辑'
```
kentcdodds: Oh, I added two things to this file last night. Could you give that a look. Thanks!

liadbiz: @kentcdodds  yes, and things need to be changed in all language versions translation....how about open an issue to let them known?

kentcdodds: Thanks @liadbiz, sorry that things are changing so quickly right now. The best way to keep things up to date is to follow the instructions in the [workflow](https://github.com/kentcdodds/glamorous-website/blob/master/other/CONTRIBUTING_DOCUMENTATION.md#workflow) section of the contributing documentation docs :)

Also, you can see when there are outdated translations for the docs in little notes under each section on the website (like on [this page](https://zh.rc.glamorous.rocks/api/))

Thanks again for all you do!

PR #231, opened 2017-07-12
exclude img.badgesize.io

tdeschryver:
**What**: Fix badge-size size on [advanced](https://rc.glamorous.rocks/advanced/)

**Why**: Because right now they are REALLY HUUUUUGE

**How**: Excluding the source `img.badgesize.io`

**Note**: Removed `markdown-wrapper.js` because it isn't being used (thought the issue was there and after some minutes I discovered this file isn't being used :frowning:).


components/markdown-wrapper.js
```diff
@@ -1,26 +0,0 @@
-import glamorous from 'glamorous'
```
kentcdodds: lol, I don't even remember this thing existing 😆 

Amazing how quickly a codebase runs away from you!

motleydev:
👍

PR #235, opened 2017-07-13
Feature/search

motleydev:
**What**: Adding style touches to #219 

**Why**: Because style makes everyone happy. And this feature rocks.

**How**: Added some new primitives to the GlobalStyles object for animation and the rest was modifying the styles/structure of the Search component.


package.json
```diff
@@ -29,7 +30,9 @@
     "prismjs": "^1.6.0",
     "prop-types": "^15.5.10",
     "react": "^15.5.4",
+    "react-autosuggest": "^9.3.1",
```
kentcdodds: Is this necessary?

package.json
```diff
@@ -29,7 +30,9 @@
...
     "react-dom": "^15.5.4",
+    "react-instantsearch": "^4.0.7",
```
kentcdodds: Is this necessary?

components/nav.js
```diff
@@ -29,6 +29,46 @@ const Navbar = glamorous.nav(({top, theme: {mediaQueries}}) => ({
...
+    outline: 'none',
+    borderColor: `${colors.primaryMed}`,
+  },
+  ':valid': {
+    maxWidth: 200,
+    minWidth: 100,
+  },
+})
+const glamorousSearch = (props, {animations}) => ({
+  ...animations.expansion('width'),
```
kentcdodds: Don't worry about the `...`, also, I don't think we'll need to use this anywhere else so I'd prefer to keep this out of `theme`. So why don't we make those as part of `inputBase`. In addition, we could also just inline that, I don't think it needs to be in a variable. Why don't we just do:

```javascript
const SearchBox = glamorous.input(
  'algolia_searchbox',
  { /* inline the styles object here */ }
)
```

components/nav.js
```diff
@@ -29,6 +29,46 @@ const Navbar = glamorous.nav(({top, theme: {mediaQueries}}) => ({
...
 }))
+// Abstract input primitives to a higher level at a later date
+const inputBase = (props, {colors}) => ({
+  width: 120,
+  borderRadius: 50,
+  paddingTop: 4,
+  paddingBottom: 4,
+  paddingLeft: 30,
+  paddingRight: 20,
+  backgroundImage: 'url(/static/images/search.svg)',
```
kentcdodds: Instead of making this external, would it be possible to just import the svg directly? We have a babel plugin that will give us a react component when we import an svg file.

components/nav.js
```diff
@@ -29,6 +29,46 @@ const Navbar = glamorous.nav(({top, theme: {mediaQueries}}) => ({
...
+const SearchBox = glamorous.input(
+  'algolia_searchbox',
+  inputBase,
+  glamorousSearch,
+)
+SearchBox.defaultProps = {
+  type: 'text',
+  pattern: '.*',
+  required: true,
+  placeholder: 'Search…',
```
kentcdodds: We'll need this to be localized :) Put it in `components/content/nav.md` and then you can do: `placeholder: content.search`

package.json
motleydev: @RIanDeLaCruz Will need to comment on this.

package.json
motleydev: @RIanDeLaCruz Will need to comment on this.

package.json
RIanDeLaCruz: @kentcdodds This is what makes docsearch run. Algolia Docsearch is a wrapper over react-autosuggest

package.json
RIanDeLaCruz: react-instantsearch can be removed already. I find it odd that it wasn't removed when I did yarn remove

components/nav.js
motleydev: We would need to restructure the search box first. It's currently just an input but to place inline graphical elements like that we'd need to do some kind of `span > input` foo and hope it doesn't break the live results function. Do you still want to do that?

RIanDeLaCruz: Pretty sure it should still be fine because docsearch just looks for the class name @motleydev

components/nav.js
motleydev: Well, this library's functional style and ease of composing were the core features of the blog post I was thinking about, but ok. :)  Done.

package.json
motleydev: I had to do some manual merges with master and this branch so it's possible I fat-fingered it.

components/nav.js
motleydev: Alright @kentcdodds - you can decide if it's still worth the server response optimization. :) I won't be able to have a go at that until this evening most likely.

components/nav.js
motleydev: Done.

motleydev:
Alright, requested changes are done except for the inline of svg - let me know!

kentcdodds:
Ok, so I've made a few changes and cleaned up the commit history. Things are a little bit broken with the algolia config and that's because [this](https://github.com/algolia/docsearch-configs/pull/18) was merged but I don't think it's been propogated yet. I think that I'll go ahead and merge this and hopefully it magically starts working when algolia reindexes things (tomorrow I'm guessing).

kentcdodds:
Thanks everyone! This was a great team effort!

giphy

motleydev:
+1 for cleaning up the merge history. This one got hairy. :)

Haroenv:
This usage of babel-plugin-preval is amazing!

kentcdodds:
😊

PR #236, opened 2017-07-14
adding dependencies

motleydev:
**What**: The website is broken, perhaps due to the dependencies not present in package.json

**Why**: Website is broken

**How**: Added dependencies into package.json


motleydev:
Not sure if this PR is necessary? When I pulled from master, the dependencies weren't there but when I compare the diff, it looks like they are?

motleydev:
I guess they are there.

PR #238, opened 2017-07-16
fix(website): fixes issue with algolia load crashing website

### LINKED ISSUE #237: Fix the website
It's broken: https://rc.glamorous.rocks/

screen shot 2017-07-15 at 10 55 59 pm

The problem fixes itself if I comment out [this algolia `docsearch` code](https://github.com/kentcdodds/glamorous-website/blob/d1ec291df19b90096e51c1e15ea3ef1974d684ed/components/algolia-config.js#L33-L41). [The error link](https://facebook.github.io/react/docs/error-decoder.html?invariant=32&args[]=34) says:

```
Unable to find element with ID 34.
```

ID 34 is the input element.

This issue is reproducible locally if you run the full build and serve the built site:

```
npm start build.en
http-server out/en
```

Not reproducible in dev mode.

Any help appreciated :smile:

luke-john:
This closes #237

**What**:

Algolia docsearch link and script moved to head to ensure availabilty

docsearch initialisation moved to componentDidMount lifecyle method of `nav.js` to ensure the element is available for it to hook onto.

**Why**:

The website was not working

**How**:

I'm not experienced with the coding style for the website, feel free to update this as you like.  Or if you give feedback happy to incorporate :).

### DIFF
--- a/__tests__/advanced.js
+++ b/__tests__/advanced.js
@@ -6,6 +6,8 @@ import Advanced from '../pages/advanced'
 expect.addSnapshotSerializer(serializer)
 expect.extend(matcher)
 
+global.docsearch = jest.fn(() => {})
+
 test('renders', () => {
   expect(() => mount(<Advanced />)).not.toThrow()
 })
--- a/__tests__/api.js
+++ b/__tests__/api.js
@@ -6,6 +6,8 @@ import Api from '../pages/api'
 expect.addSnapshotSerializer(serializer)
 expect.extend(matcher)
 
+global.docsearch = jest.fn(() => {})
+
 test('renders', () => {
   expect(() => mount(<Api />)).not.toThrow()
 })
--- a/__tests__/basics.js
+++ b/__tests__/basics.js
@@ -6,6 +6,8 @@ import Basics from '../pages/basics'
 expect.addSnapshotSerializer(serializer)
 expect.extend(matcher)
 
+global.docsearch = jest.fn(() => {})
+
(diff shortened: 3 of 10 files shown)


pages/index/index.js
```diff
@@ -3,6 +3,7 @@ import glamorous from 'glamorous'
 import {LiveProvider, LiveEditor, LiveError} from 'react-live'
 import {LiveContextTypes} from 'react-live/lib/components/Live/LiveProvider'
 import Head from 'next/head'
+import {AlgoliaLink, AlgoliaScript} from '../../components/algolia-config'
```
kentcdodds: Could you put these changes in the layout component? That way it's on all pages. Thanks!

luke-john: Ahh, just tried that and it looks like the next Head component doesn't like multiple Heads.

One option could be to change the shape of how Layout is used to instead have a render method which passes PageHead as a prop

```
// Layout.js
   const PageHead = () => (
        <Head>
            <AlgoliaLink />
            <AlgoliaScript />
            {children}
        </Head>
    )
   ...
    <Div overflow="auto" width="100%">
        {renderChildren({PageHead})}
    ...

// Page.js
<Layout
    render={({PageHead}) => {
        ...
            <PageHead>
                ...
```

happy for other suggestions though

luke-john: After looking in the docs and not seeing anything, I tried again and it seems all fine with multiple head components. have updated :)

PR #252, opened 2017-07-28
remove rc!

kentcdodds:
**What**: Remove the rc

**Why**: Because we're getting ready to release this and I want the website ready before we announce it.

**How**: Remove the `rc.` from the `now-travis` file

### DIFF
--- a/other/now-travis
+++ b/other/now-travis
@@ -107,8 +107,7 @@ async function deploy(context, sha) {
     // if we're production, then we need to alias things
     console.log(`↪️ We're building for production, so let's alias this!`)
     const prefix = locale === fallbackLocale ? '' : `${locale}.`
-    // TODO: remove the `rc` when we're ready to go live
-    const domain = 'rc.glamorous.rocks'
+    const domain =  'glamorous.rocks'
     const alias = `${prefix}${domain}`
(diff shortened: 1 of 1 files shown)


tdeschryver:
Seems like every `process.env` which is being used is broken at run time? :thinking: 
- `process.env.LOCALE`: current language in `locale-chooser`
- `process.env.LOCALE`: Twitter card prefix
- `process.env.LOCALE`: `components\utils\locale`
- `process.env.USE_PREFETCH`: in nav anchor tags

kentcdodds:
Yeah, I just noticed that as well... It's weird because I added the babel plugin which should transpile that... Could you try to give that a look? If we have to I'm happy to add the webpack define plugin back in...

tdeschryver:
Think the solution would be to use `[babel-plugin-transform-inline-environment-variables](https://babeljs.io/docs/plugins/transform-inline-environment-variables/)` instead of `babel-plugin-transform-node-env-inline`. Let me run some more tests to be completely sure.

kentcdodds:
Sounds great! Could you put up a PR for that?

tdeschryver:
Of course! :smile:

PR #253, opened 2017-07-28
Transpile process env

tdeschryver:
**What**: Transpile `process.env`
(Also took the opportunity to refactor the use of the current locale, see 9ecbe8aaf1f76cb314394c54a33cfff3157ea0a2)

**Why**: `LOCALE` and `USE_PREFETCH` are `undefined` at the moment

**How**: By using [`babel-plugin-transform-inline-environment-variables`](https://babeljs.io/docs/plugins/transform-inline-environment-variables/) instead of `babel-plugin-transform-node-env-inline`.

### DIFF
--- a/components/utils/locale.js
+++ b/components/utils/locale.js
@@ -1,7 +1,12 @@
 const {supportedLocales, fallbackLocale} = require('../../config.json')
 
-let {LOCALE: lang} = process.env
-if (!supportedLocales.includes(lang)) {
-  lang = fallbackLocale
+function locale() {
+  const {LOCALE: lang} = process.env
+  if (supportedLocales.includes(lang)) {
+    return lang
+  }
+
+  return fallbackLocale
 }
-module.exports = lang
+
+module.exports = () => locale()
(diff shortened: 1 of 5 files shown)


components/utils/locale.js
```diff
@@ -1,7 +1,12 @@
...
+  const {LOCALE: lang} = process.env
+  if (supportedLocales.includes(lang)) {
+    return lang
+  }
+
+  return fallbackLocale
 }
-module.exports = lang
+
+module.exports = () => locale()
```
kentcdodds: Why export the function? Couldn't we just export the lang?

tdeschryver: Because when we're running our tests we're setting `process.env.LOCALE` a couple of times and it will keep using the first set value in our component, which will cause our tests to fail.
At least that is what I'm thinking...

Maybe there is a better way of doing this, but I don't know how :thinking: .

kentcdodds: Meh, it's ok, we'll go with this for now. Thanks!

PR #255, opened 2017-07-31
fix(docsearch): execute the language for docsearch

Haroenv:
**What**: docsearch initialisation

**Why**: because no results were showing

**How**: the lang function wasn't executed before being evaluated

There is still something weird, so I'll probably add a second commit that fixes that too


Haroenv:
@kentcdodds, it works perfectly now!

Haroenv:
perfect 👌

PR #259, opened 2017-08-04
prepaired things for russian translation

IgorKonovalov:
Preparations for russian translation for glamorous website 

Because russian natives need to have access to documentation

By translating pages


components/locale-chooser.js
```diff
@@ -234,6 +235,11 @@ function mapLocale(key = fallbackLocale) {
       display: '中文',
       Flag: CnSvg,
     },
+    ru: {
+      key,
+      display: 'Russian',
```
kentcdodds: Oh, actually, this should be translated into Russian. Could you update that?

PR #271, opened 2017-08-19
breadcrumbs example

kristian-puccio:
**What**: Added example showing Breadcrubs

**Why**: More examples of using 'complex' selectors.

**How**: Copy other examples


kristian-puccio:
Ok I think that's fixed.

Cheers

PR #275, opened 2017-08-31
filterProps API docu

bjoernricks:
First draft of the filterProps api usage. Feel free to change wording, examples, etc.


pages/api/content/glamorous.md
```diff
@@ -106,6 +106,64 @@ const MyStyledComponent = glamorous(MyComponent, {
...
+
+<MyComponent flex />
+// this will render a <div> element with the display CSS property set to 'flex'.
+// The withFlex HOC can't set a `rootEl` because it accepts all kind of components.
+// The flex property will not be forwarded to the div because it is listed in
+// `filterProps`.
+```
+
+Setting a property in `filterProps` always takes precedence over `forwardProps`.
+If the same property is contained in both arrays it will always be filtered.
```
kentcdodds: Honestly, I'm not sure how valuable this bit is. Seems weird for anyone to even try 🤷‍♂️

bjoernricks: You are right. Nobody would do that by intention. I'll remove that paragraph.

PR #276, opened 2017-09-03
fix(typescript): update typescript documentation

luke-john:
**What**:

Updates to the documentation for the typescript defintions 

**Why**:

The documentation for the definitions were out of date and contained incorrect information.


kentcdodds:
Thank you! As always!

Issue #283, opened 2017-10-20
Failing tests on newly pulled down repo


andrew-codes:
**Problem Description:**
When running `npm start validate` on a freshly pulled down repo yields failing tests. Running the `npm start test` command shows failed tests with a more descriptive error message.

`npm start validate` CLI output
![2017-10-20_13-28-59](https://user-images.githubusercontent.com/1578436/31834177-204da80a-b59b-11e7-9b91-c3026223ae06.png)

**Environment**
- OSX
- node@8.7.0
- npm@5.4.2

**Suggested Solution:**
Speaking to others online, it seems not everyone has this issue. Due to this, I suspect it may be environmental; possibly relying on a globally installed package or having a globally installed package that interacts in some way.

kentcdodds:
Hmmm.. Yeah, that's weird. It looks like `babel-macros` is failing to be included in the transforms babel's using when running the tests. I'm not sure why though... I'll come back to this if I ever have time 😢

andrew-codes:
Ok, I will investigate further. I wanted to be sure I didn't miss a step but sounds like it could be an issue. I'll dig a bit deeper and see what I can find.

kentcdodds:
Thank you @andrew-codes!

Shayla303:
I had the same issue pulling the repo down. You  do npm install and then do an npm install node as well. It updates the dependencies calling on the older version of node. I am on a mac and this solution worked for me. 
<img width="657" alt="solution" src="https://user-images.githubusercontent.com/24359681/35783318-5405d2d2-09c2-11e8-9dc0-ae68a9c1c0a8.png">

I have issues running the start with nps, which I will work on figuring out as well.

kentcdodds:
This project is now in an unmaintained status. Please see [the glamorous README](https://github.com/paypal/glamorous) for more information.

PR #285, opened 2017-11-03
Update prop styles example

peterschussheim:
**What**:

- [x] update codesandbox example and add link to `prop-styles`
- [x] bump `babel-macros` and `nps version`: Referencing #283, I had this issue too and fixed it with version updates.  I am not sure if this fixes the problem for all users.
- [x] update .gitignore and commit package-lock

**Why**:

- I'd like to expose more `glamorous` users to the `prop-styles` utility

**How**:

[View deployed build](https://en-glamorous-xspzqgxrxi.now.sh).


peterschussheim:
Failing on Travis due to:

```
The macro imported from "preval.macro" must be wrapped in "createMacro" which you can get from "babel-macros".
```

Could this be due to the version bump?  All local tests pass:
iterm2001

kentcdodds:
Yeah, there's an old version of something. I've cleared the travis cache. We'll see what happens with the new build.

kentcdodds:
Oh, I just realized what's going on. You used `npm` to install dependencies. But this repo uses `yarn`. If you could remove the `package-lock.json` file and then use `yarn` to install dependencies that would fix it (it'll update the `yarn.lock` file). Also, you'll need to pull my changes before you do any of this. Thanks!

P.S. I really don't like committing lockfiles in open source projects... But that's what we're doing in this one so we'll just have to do that.

peterschussheim:
awesome! 👍

PR #289, opened 2018-02-09
Update glamorous-component-factory.md example code

sgnl:
{"<div class="<glamor-generated-class> other-class">content</div>" => "<div class="<glamor-generated-class other-class">content</div>"}

notice the removal of `>` after the text `glamor-generated-class`

Notice this typo while reading the documentation @ https://www.glamorous.rocks

**What**:

**Why**:

**How**:


sgnl:
You are correct!

